PR #63, opened 2026-06-28
feat(diarization): isolate pyannote in a CPU-torch sidecar subprocess

Artexis10:
## Why

`KB_MCP_DIARIZE` diarization was **enrollment-ready but pipeline-blocked** on the live box. The ECAPA named-attribution layer works, but the pyannote *who-spoke-when* pipeline cannot load: the service runs a custom **torch-2.12+cu132** (Blackwell) build, and pyannote + torchaudio are fundamentally incompatible with it — **six version walls** (torchcodec native-lib load failure; `torchaudio.AudioMetaData` + `list_audio_backends` removed in torchaudio 2.11; speechbrain 1.x `LazyModule`; `huggingface_hub` removed `use_auth_token`; the `use_auth_token`→`token` kwarg rename). These are **version walls, not GPU walls** — the cu132 pin transitively forces the exact broken combo, and CPU torch alone doesn't fix them.

## What

Run **only** the pyannote pipeline in an **isolated CPU-torch sidecar venv** (`sidecar/diarizer/`) as a subprocess, pinned to the canonical pyannote-3.1 stack where every wall is absent. The seam is tiny: `extract._run_diarization` is the only function that touched pyannote, so it becomes a thin subprocess call (audio path in → JSON turns out via an out-file). Everything downstream — ECAPA voice-profile attribution, transcript assembly — is **unchanged** and stays in the main venv.

- **`sidecar/diarizer/`** — standalone uv project pinning `pyannote.audio 3.1.1` / `torch 2.2.2+cpu` / `torchaudio 2.2.2+cpu` / `speechbrain 0.5.16` / `huggingface_hub 0.25` (Python 3.12, no torchcodec). `worker.py` decodes via faster-whisper's `decode_audio` (timebase parity with the main ASR), redirects stdout→stderr, and writes `{"turns":[…]}` to an out-file. **Pins are load-bearing** — `uv lock --upgrade` would float back into the trap.
- **`extract._run_diarization`** — subprocess seam: locate the sidecar python (`KB_MCP_DIARIZE_SIDECAR_PYTHON` override), spawn with a duration-scaled timeout, `CUDA_VISIBLE_DEVICES=""` + merged env, parse the out-file. **Soft-fail preserved byte-for-byte**: sidecar absent / nonzero exit / timeout / bad output → plain transcript, never raises. The dead in-process pipeline symbols are deleted.
- **Main `pyproject.toml`/`uv.lock`** — `pyannote.audio` removed from the `diarization` extra (keeps `speechbrain` for ECAPA). This drops `torchcodec` from the main venv and retires the per-`uv sync` `uv pip uninstall torchcodec` dance + a latent hazard to the bge/CLIP embedding stack.
- **`scripts/setup-diarizer.ps1`** — per-box provisioning (deploy-time only; `uv` is never needed at service runtime). README env + provisioning docs.
- **OpenSpec** change `diarization-cpu-sidecar` (validated `--strict`).

## Verification

- **Empirically validated on the Windows/cu132 box**: the full sidecar stack imports (`AudioMetaData`, `list_audio_backends`, `Pipeline`, `decode_audio` all load) and `worker.py` runs end-to-end on a synthetic WAV — decodes, reaches `Pipeline.from_pretrained`, and fails *cleanly* only at the gated model (no token). All seven walls cleared.
- **Full suite green**: 824 passed. `ruff` clean. New hermetic soft-fail tests (`test_diarizer_sidecar.py`, subprocess mocked) + a gated real-worker smoke (`test_diarizer_worker_smoke.py`, skips when the sidecar venv is absent).
- **Remaining (deploy box, @Artexis10)**: `setup-diarizer.ps1 -Prewarm` + `HUGGINGFACE_TOKEN` (accept `speaker-diarization-3.1` + `segmentation-3.0`) → `KB_MCP_DIARIZE=1` → restart → diarize a real 2-speaker recording → confirm `[Hugo]: …` / `[Speaker B]: …` render and the embedding stack is intact.

https://claude.ai/code/session_01DnHCrSGB62kFmmptCoMAQn

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ the repo root). The only required one is the vault path.
 | `KB_MCP_IMAGE_TAGS_TOPK` | Max image tags to emit per image (default `5`). |
 | `KB_MCP_IMAGE_TAGS_THRESHOLD` | Raw-cosine floor a tag must clear (default `0.22`). |
 | `KB_MCP_DIARIZE` | Set to enable opt-in ASR speaker diarization (`[Speaker A]: …` turns). Requires the diarizer sidecar (see below). |
+| `KB_MCP_DIARIZE_DEVICE` | Sidecar device: `cpu`/`cuda`/`auto` (default `auto` → GPU when available, else CPU). |
 | `KB_MCP_DIARIZE_SIDECAR_PYTHON` | Override path to the diarizer sidecar's Python (default `sidecar/diarizer/.venv/Scripts/python.exe`). |
 | `KB_MCP_DIARIZE_TIMEOUT` | Seconds the sidecar subprocess may run before soft-failing to a plain transcript (default: `max(900, duration×6)`). |
 | `KB_MCP_DIARIZE_MODEL` | pyannote checkpoint the sidecar loads (default `pyannote/speaker-diarization-3.1`). |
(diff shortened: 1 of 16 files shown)


Artexis10:
## Update: the sidecar now runs on the GPU (cu130), ~23× faster

The initial CPU pin was over-conservative — Q runs the *same* pyannote diarization on the *same* Blackwell RTX 5080 in production, which proved CPU was unnecessary. Pivoted the sidecar to Q's proven stack:

- **pyannote.audio 4.x** (loads the `speaker-diarization-3.1` model) + **torch 2.9.1+cu130** (Blackwell `sm_120`) + **torchaudio 2.9.1**. The incompatible build is the main service's bleeding-edge `torch 2.12+cu132` (it forces torchaudio 2.11, which dropped the APIs pyannote needs); `2.9.1+cu130` is the sweet spot — `sm_120` kernels **and** pyannote compatibility. pyannote **4.x** package is required because the 3.1 package calls the removed `torchaudio.set_audio_backend`; 4.x loads the faster 3.1 model and returns a `DiarizeOutput` the worker unwraps. torchcodec rides along but is unused (we pre-decode → waveform dict).
- New `KB_MCP_DIARIZE_DEVICE` (cpu|cuda|auto, default auto). `_run_diarization` keeps the GPU visible for cuda/auto and **strips the cu12 `nvidia/*/bin` dirs from the child PATH** so they can't shadow the sidecar's cu130 cuDNN (mirrors the CLIP/cuDNN-shadow fix).
- **Verified end-to-end on the RTX 5080: 47s vs 18min CPU** for a 37-min track (~23×); a 3-person mix diarizes to multiple speakers. Full suite 828 passed, ruff clean, `openspec validate --strict` passes.

Note: the branch/dir name still says "cpu-sidecar" (cosmetic only) — it's GPU-capable with a CPU fallback.

https://claude.ai/code/session_01DnHCrSGB62kFmmptCoMAQn

PR #125, opened 2026-07-05
perf(embeddings): key the matrix caches on a write generation, not the WAL-checkpoint mtime

Artexis10:
## What

Key the embedding + CLIP matrix caches (and find's hot cache) on an **in-band write generation** instead of the sidecar file's mtime. This is PR2 of the "keep real-vault find() ~280ms under active editing" arc.

## The proven WAL-checkpoint mechanism (root cause)

The `.embeddings.sqlite` / `.clip.sqlite` sidecars run in WAL mode. A **commit does not move the main file's mtime — a checkpoint does.** Under any concurrent connection holding a read txn, SQLite defers the checkpoint to whichever connection closes *last* (often a pure reader), at a moment no `_patch_cache` runs. Two consequences on the old mtime-keyed caches:

- **Spurious full reloads:** `_patch_cache` recorded a pre-checkpoint mtime; the next `all_vectors()` saw the post-checkpoint mtime and full-reloaded a matrix that was already correct (~397ms idle, seconds under load).
- **`find._freshness_key` staleness in both directions:** keying the hot cache on the sidecar `st_mtime_ns` gave spurious misses *and* a stale-hit path (an uncheckpointed commit leaves the mtime unmoved → the cache serves stale results).

Single-threaded unit tests never caught this — a solo connection always checkpoints consistently. The race only appears under the reader/writer concurrency of the live service.

## The generation design

- **`meta(key TEXT PRIMARY KEY, value INTEGER)`** created in `_connect()`. `generation` is bumped **inside each write's own transaction** (`upsert_file` / `delete_file` / `upsert` / `upsert_frames` / `delete`, and `rebuild_all`'s **final bulk txn only**, never the wipe) via `UPDATE … value+1` → `INSERT` if absent → read-back, all under the write lock. A separate **`epoch`** key, bumped only by `rebuild_all`, catches re-embeds that change no file mtimes.
- **`all_vectors()`** drops the stat. It reads `(epoch, gen)` per call and serves the existing lock-free snapshot when equal; a mismatch full-reloads capturing `gen`+`epoch`+rows in **one explicit `BEGIN` snapshot** (bare SELECTs in autocommit are each their own snapshot).
- **`_patch_cache`** takes the writer's own in-txn generation and advances the cached generation **only when contiguous** (`own == cached + 1`) — never `max()`, which would let the cache claim a generation whose rows it lacks if a writer died between commit and patch.
- **`find._freshness_key`** keys on new `EmbeddingIndex.cache_token()` / `ClipIndex.cache_token()` = `(epoch, generation)` (precedent: `lexstore.cache_token`).
- **Version-skew fallback:** a legacy sidecar whose generation reads 0 (pre-meta) keeps today's mtime keying until the first gen-bumping write; the `meta` table is migrated in on first connect (one-time log).
- **Observability:** every full matrix load logs `reason=cold|epoch|legacy|genuine` — the blind spot that made this bug cost two debugging sessions, now greppable in `logs/exomem.log`.

Third occurrence of the WAL-mtime class in this repo (precedent: `lexstore.cache_token`). `claims.py` has the same pattern — **documented follow-up, not touched here.** The vec0 KNN ladder in `search()` is untouched and still serves without loading the matrix (invariant asserted in `tests/test_vecstore.py`).

## Test evidence (TDD, red-first)

The proving test `test_utime_bump_alone_does_not_reload` was **RED on the old code** (`assert 1 == 0` — a bare mtime bump forced one spurious full reload), and is green now. New/updated tests in `tests/test_embedding_matrix_cache.py` + `tests/test_find_hot_cache.py`:

- utime-only mtime bump → no reload
- writer commits under a held reader txn (reader closes last) → no reload — the live race in test form
- external second-instance write / delete → detected via generation, new rows served
- external `rebuild_all` with a **frozen** sidecar mtime → fresh vectors served (epoch), plus an epoch-only isolation test
- legacy sidecar (no `meta`) migrates + gen==0 mtime fallback still invalidates
- out-of-order patch (gen gap) → cached generation does not advance past the gap; next `all_vectors()` converges
- ClipIndex mirrors (utime no-reload, external write, external delete)
- find hot-cache invalidates via a gen-bumping write (rewritten from the old utime test)

## Gates

- `uv run pytest` (full lean suite): **1557 passed, 3 skipped**
- `uv run --extra embeddings python -m pytest tests/test_retrieval_golden.py -m embeddings`: **3 passed** (recall floors: ndcg10 ≥ 0.85, mrr ≥ 0.80, recall10 ≥ 0.88)
- `uvx ruff check` on the changed files: **zero new errors** (test files clean; source files carry only pre-existing advisory findings)
- vec0/CLIP/video/hybrid suites (with embeddings extra): **72 passed** — the "vec0 never loads the matrix" invariant holds

https://claude.ai/code/session_01CEsYrCghxuK4yCaC8QoMUV

### DIFF
--- a/src/exomem/audit.py
+++ b/src/exomem/audit.py
@@ -1408,7 +1408,7 @@ def _check_corpus_contradictions(
     supersede; nothing is ever auto-acted.
 
     Reuses the existing vector sidecar (`EmbeddingIndex.all_vectors()`, cached by
-    mtime) — it reads the chunk vectors already on disk and never re-encodes, so the
+    the write generation) — it reads the chunk vectors already on disk and never re-encodes, so the
     sweep is O(eligible_files) matmuls over the sidecar matrix and needs no model
     (works on a CPU/offline box as long as a sidecar exists). The band edges are the
     same knobs the write-time check uses: floor = `corpus_aware._contradiction_floor`,
@@ -1450,7 +1450,7 @@ def _check_corpus_contradictions(
         return []
 
     idx = embeddings_module.get_embedding_index(vault_root)
-    metadata, matrix = idx.all_vectors()  # cached by sidecar mtime
+    metadata, matrix = idx.all_vectors()  # cached by the sidecar write generation
     if not metadata or matrix.shape[0] == 0:
         return []
 
--- a/src/exomem/embeddings.py
+++ b/src/exomem/embeddings.py
@@ -878,6 +878,116 @@ def _file_block(keys: list[str], rel_path: str) -> tuple[int, int]:
     return ins, ins
 
 
+# --------------------------------------------------------------- generation meta
+#
+# The matrix caches key on an in-band WRITE GENERATION, not the sidecar file's
+# mtime. The sidecars are WAL sqlite: a commit does NOT move the main file's
+# mtime — a CHECKPOINT does, and under concurrent connections the checkpoint
+# fires whenever the last connection (often a pure reader) closes, at a moment no
+# writer runs. So mtime-keyed invalidation BOTH spuriously misses (a checkpoint
+# with no content change) AND goes stale (an uncheckpointed commit leaves the
+# mtime unmoved). A `meta(key, value)` row bumped inside each write's own
+# transaction changes iff the content did. Third occurrence of this class in the
+# repo; precedent + rationale: lexstore.cache_token.
+
+
+def _ensure_meta_table(
+    conn: sqlite3.Connection, data_table: str, sidecar_name: str
+) -> None:
+    """Create the `meta(key, value)` generation table if absent (idempotent).
+
+    Logs ONCE when the table is first created over an EXISTING non-empty sidecar
+    (a legacy DB written before this binary): its generation reads 0 until the
+    first gen-bumping write, and the caches fall back to mtime keying meanwhile.
+    """
+    existed = (
+        conn.execute(
+            "SELECT 1 FROM sqlite_master WHERE type='table' AND name='meta'"
+        ).fetchone()
+        is not None
+    )
+    conn.execute("CREATE TABLE IF NOT EXISTS meta (key TEXT PRIMARY KEY, value INTEGER)")
+    if not existed:
+        # data_table is a trusted literal ("chunks"/"images"), never user input.
+        has_rows = (
+            conn.execute(f"SELECT 1 FROM {data_table} LIMIT 1").fetchone() is not None
+        )
+        if has_rows:
+            log.info(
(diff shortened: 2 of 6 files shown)


Artexis10:
Pushed a follow-up commit addressing the 5 review findings (F1-F5):

**F1 (MUST, correctness)** — `_patch_cache` no longer splices content unconditionally. Content and the label now advance under the SAME gate: `own_epoch == cached.epoch AND own_instance == cached.instance AND own_gen == cached.generation + 1`. On any mismatch the splice is skipped entirely (no content change, no label move); the resulting token mismatch heals via a full reload on the next `all_vectors()`. Two new tests reproduce the exact corruption traces: a delayed same-file patch after a newer write lands, and a delayed patch racing `rebuild_all`'s epoch bump — both assert the stale rows never reach the served matrix.

**F2 (MUST, robustness)** — the token read on the fast path is now wrapped: a locked/corrupt sidecar no longer raises out of `all_vectors()` (it did, which would have taken down `audit.py`'s all-pairs sweep). A warm cache is served through the failure; only a cold cache falls through to the full-load attempt. New test monkeypatches `sqlite3.connect` to raise and asserts the warm cache is served without an exception propagating.

**F3 (hardening)** — added a random nonzero `instance` id, written once when the `meta` table is first created, as a third component of every token read/comparison. Guards the ABA case where a sidecar is deleted and recreated from scratch (restarting `epoch`/`generation` at 0) and could otherwise coincidentally match a still-warm cache from the old file. New test deletes+recreates a sidecar, confirms `(epoch, gen)` alone would have collided, and confirms the instance mismatch catches it.

**F4 (dedupe + cost)** — hoisted `_read_token`/`_cache_matches` off both classes into shared module-level helpers; the per-call freshness peek now uses the same bare-connection pattern the public `cache_token()` already used (no WAL pragmas, no schema DDL — this read happens 4-6x per query). Converted the positional cache tuples to two small NamedTuples (`_EmbCache`/`_ClipCache`) now that they carry four token fields instead of two.

**F5 (docs)** — corrected the `rebuild_all` comment that overclaimed a cold reader's behavior in the wipe→final-txn gap (it loads empty rows under the pre-bump token, same exposure a full reload always had — not a new guarantee). Documented the one-way legacy-sidecar fallback explicitly (old/new binaries can't safely coexist past the first gen-bumping write; fine for this single-user deployment).

Gates after the fix: full suite 1572 passed / 3 skipped (pre-existing, unrelated — diarizer sidecar venv, missing `av`); golden gate 3 passed; vec0/CLIP/video/hybrid/audit suite 89 passed; ruff clean on all changed files.

PR #144, opened 2026-07-07
Benchmark Rust paths and trim write latency

Artexis10:
﻿## Summary

- Add Rust comparator binaries and Python harnesses for keyword find and write-path benchmarks.
- Bound keyword-mode and hybrid keyword exact-match work so find does not scan unbounded sets on large vaults.
- Trim write latency by skipping resolver/index work that is unnecessary for plain edits and navigation files.

## Write benchmark results

Synthetic 2,000-page vault after fixes:

- production edit: 31.6 ms median
- Python file-only edit: 1.5 ms median
- Rust file-only edit: 1.0 ms median
- production note: 78.9 ms median
- Python/Rust file-only note create: about 1 ms median

Synthetic 10,000-page edit benchmark after fixes:

- production edit: 32.0 ms median
- Python file-only edit: 1.6 ms median
- Rust file-only edit: 1.1 ms median

Markdown/YAML-only scratch copy of the real vault corpus, no live-vault mutation:

- production edit: 47.6 ms median
- Python file-only edit: 1.4 ms median
- Rust file-only edit: 0.9 ms median
- production note: 211.9 ms median

Conclusion: the file-mutation slice is already low-ms in Python; the latency is from production side effects, especially subindex/count refresh and lexical sidecar/log maintenance. This does not support a project rewrite in Rust.

## Tests

- `PYTHONPATH=src .venv\\Scripts\\python.exe -m pytest -q -p no:cacheprovider`
- Result: 1566 passed, 16 skipped


experiments/rust_find_keyword_lite/src/bin/rust_write_lite.rs
```diff
@@ -0,0 +1,349 @@
...
+    title: String,
+    note_type: String,
+    content: String,
+    date: String,
+}
+
+fn main() -> ExitCode {
+    match run() {
+        Ok(()) => ExitCode::SUCCESS,
+        Err(err) => {
```
pxp9, line 122: it would be nice to have different exit codes depending on the error

experiments/rust_find_keyword_lite/src/bin/rust_write_lite.rs
```diff
@@ -0,0 +1,349 @@
...
+                i += 1;
+                vault = args.get(i).map(PathBuf::from);
+            }
+            "--path" => {
+                i += 1;
+                path = args.get(i).cloned();
+            }
+            "--old" => {
+                i += 1;
+                old = args.get(i).cloned();
```
pxp9: these cloned should waste a lot of time requesting heap memory to the system, i think you should take clap approach which it uses &str references

Artexis10:
﻿Addressed this in follow-up commit 07eee5c. The Rust comparators now use clap derive for CLI parsing and serde/serde_json for output encoding, so the hand-rolled argument parser and manual JSON escaping are gone. I disabled clap default features to avoid pulling in the Windows terminal color stack (`windows-sys`/`dlltool`) on this toolchain; release build still passes.

Verified:
- `cargo build --release --manifest-path experiments/rust_find_keyword_lite/Cargo.toml`
- `PYTHONPATH=src .venv\Scripts\python.exe -m pytest -q tests\test_rust_keyword_lite.py -p no:cacheprovider`
- smoke: `PYTHONPATH=src .venv\Scripts\python.exe scripts\compare_rust_write_ops.py --samples 3 --pages 200`

experiments/rust_find_keyword_lite/src/bin/rust_write_lite.rs
```diff
@@ -0,0 +1,248 @@
...
+    #[arg(long)]
+    path: String,
+    #[arg(long)]
+    title: String,
+    #[arg(long = "note-type", default_value = "insight")]
+    note_type: String,
+    #[arg(long)]
+    content: String,
+    #[arg(long, default_value = "2026-07-07")]
+    date: String,
```
pxp9, line 108: i think you dont need owned String , try to change to &str reference

experiments/rust_find_keyword_lite/src/main.rs
```diff
@@ -0,0 +1,276 @@
...
+
+fn run() -> Result<(), String> {
+    let args = Args::parse();
+    let started = Instant::now();
+    let query_norm = args.query.trim().to_lowercase();
+    let tokens: Vec<&str> = query_norm.split_whitespace().collect();
+    let mut stats = Stats::default();
+    let mut hits = Vec::new();
+
+    if !tokens.is_empty() {
```
pxp9: could you invert this condition ?

if it is empty return an error or return Ok(())

Artexis10:
﻿Addressed the new review points in commit 6f64313.

Changes:
- `rust_write_lite` now returns distinct runtime exit codes:
  - 2 usage/input
  - 3 data/no-match
  - 4 conflict/ambiguous/already-exists
  - 5 I/O
  - 6 output encoding
- Added a pytest assertion for the new exit-code behavior.
- Inverted the empty-query branch in `rust_find_keyword_lite` so it early-returns an empty result instead of nesting the main scan.
- Moved the `rust_write_lite` internal timer to after `Cli::parse()`, so `rust_file_lite` measures the write slice and `rust_process_wall` captures CLI/startup overhead.

On the `String` vs `&str` point: with `clap` derive, the parsed CLI struct owns argv values (`String`/`PathBuf`) because those values need storage after parsing. The hot code borrows from those owned fields (`&args.path`, `&args.old`, etc.). Making the parsed struct fields `&str` is not a practical derive-path win here; it either complicates lifetimes around clap's matches storage or just moves ownership elsewhere. The benchmark conclusion does not depend on this allocation cost because the file-operation timer now starts after parse, and process wall time reports parse/startup separately.

Verified:
- `cargo build --release --manifest-path experiments/rust_find_keyword_lite/Cargo.toml`
- `PYTHONPATH=src .venv\Scripts\python.exe -m pytest -q tests\test_rust_keyword_lite.py -p no:cacheprovider`
- smoke: `PYTHONPATH=src .venv\Scripts\python.exe scripts\compare_rust_write_ops.py --samples 3 --pages 200`

Artexis10:
Closing this old draft as an obsolete merge vehicle: it is 418 commits behind and conflicts with the evolved find/lexstore/write APIs. The useful benchmark conclusion is preserved; any desired optimization should be remeasured and proposed as a narrow current-main change. Branch and history remain available.

PR #234, opened 2026-07-15
feat: Adoption Studio v1 — governed adoption runs, agent proposals, Studio UI, hosted staging

Artexis10:
Implements OpenSpec change `add-adoption-studio` (proposal/design/specs/tasks in this branch).

## What

- One governed product command `adoption_studio` (ten actions) — MCP + REST + CLI + hosted gateway from a single `_PRODUCT_SPEC` entry; `status`/`work-item` read-only classified.
- Durable canonical-file run store under `Knowledge Base/_Adoption/runs/<id>/` via `batch_atomic_write`; explicit state machine with preview-exact-actions (`plan_id` + `selection_hash`), write-time re-validation, per-item outcomes, failed-subset retry, interrupted-apply recovery.
- AdoptionWorkItem agent contract: five proposal kinds applied only through existing governed leaves; proposals surface as `exomem://review/adoption/<id>` items through `review_memory`/`triage_memory`/`review_item_context`; zero-arg `continue_adoption` MCP prompt + per-run MCP resource.
- Studio UI `?view=adopt`: nine guided screens, persistent originals-untouched guarantee, honest verification line from real re-hash counts (assets bumped to v4 shell).
- Thin hosted entrypoint: upload/ZIP staging under `_Staging/adoption/<run_id>/` with zip-slip protection and caps; admission via generic dispatch, no intercept-set changes, `writer_lease.py` untouched.

## Verification

- 2952 passed / 0 failed lean suite (70 new adoption tests), latency + golden-schema + scaffold-no-leak gates green; `openspec validate --all --strict` 73/73; ruff clean.
- Browser acceptance: 8 new Playwright scenarios for the adoption journey, 11/11 lane green.
- V.2 deterministic-only journey (no assistant): `recall_check.ok`, originals byte-identical.
- V.1 real-vault journey through the served UI (no mocks): 5 notes imported with provenance, originals proven byte-identical by external sha256 snapshot. This journey exposed a UI↔engine selection contract bug (default-ON selection materialized empty server-side) fixed in `46313d9` with red-first engine tests and a spec scenario.

Remaining before archive: confirm the Substrate Home integration spec handoff (V.4, design.md §Substrate Home).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YNKAiTv1sKMCgCDnZ73qSQ

PR #280, opened 2026-07-20
feat: integrate Adoption Studio with semantic units (packs, propose-time contract findings, Studio surface)

Artexis10:
## What

Integrates Adoption Studio with the semantic-units layer (OpenSpec change `integrate-adoption-semantic-units`): adoption work items now carry semantic-unit context packs per source, proposals are validated against the semantic write contract at **propose time** (not just at apply), and the Studio surfaces those findings to the reviewer before approval.

- **Lane A — engine** (34937f6, `adoption_proposals.py`):
  - `work_item` source rows gain `semantic_units`, built by the same pack constructor the primary context-pack surface uses (reused, not reimplemented), bounded by the clamped `max_chars_per_source`.
  - `propose` runs the create path's `validate_only` for compilation/supersession with the payload mapped exactly as `_route_apply` maps it; persists compact `contract_findings` (code/severity/detail) plus `reviewed_none_required`/`committable_after_review`; non-review blockers classify the proposal `invalid` via the existing findings mechanism. The validate call writes nothing (byte-snapshot-asserted).
  - `_item_view` and `assemble_context` expose the findings so the Studio needs no extra call.
- **Lane B — Studio** (0b3d525, studio assets, render-only):
  - `contractFindingsView` (pure, node-eval-tested) turns findings into display lines, states the fixed reviewed-none consequence sentence, and flags approve-disabled for server-marked `invalid` proposals; `openProposalDetail` renders it with the failure-group styling and suppresses the approve control (honesty — the server refuses invalid applies regardless).
  - Fixes the `junkTotal` falsy-zero bug via extracted `junkCount` (a present-but-zero `junk_counts` map is authoritative).
- **Review fix round** (e15982a, after two independent finder passes):
  - `_contract_validation` no longer labels every `validate_only` exception CONTRACT_BLOCKED: only a genuine `SEMANTIC_CONTRACT_BLOCKED` keeps that framing; other errors (e.g. PROJECT_KEY_TYPO — the note path flattens `NoteError` into a plain `ValueError` with the code only as a `str()` prefix) surface as `VALIDATION_FAILED` with the true error text.
  - `contractFindingsView` falls back to the context's generic `findings` (then one fixed generic line) when an `invalid` proposal has no contract findings, so a disabled approve is never unexplained. Unreachable via the queue today (`build_queue` filters invalid proposals) — belt-and-braces honesty.

Reviewed-none is spec'd, not changed: a regression test pins that the review reason equals the approver's `why`.

## Verification

- `openspec validate --all --strict`: 78/78.
- Full `pytest -q` on final HEAD: **4046 passed, 61 skipped**; the 54 failures (53 `test_install_hook.py`, 1 socket-preflight in `test_hosted_binding_v2.py`) reproduce **identically on clean main** in the same sandbox — environmental (sandbox tmp ownership / socket denial), not from this change.
- Browser chromium suite: **12/12**, including the new invalid-proposal test (adversarial shape: empty `contract_findings`, populated `findings`) and the extended suggestions test.
- ruff clean on changed Python; golden tool-schema fixture byte-identical vs origin/main (no command signature changes).
- Unmocked V.2 journey on a real vault: propose a compilation without typed relations → `RELATION_DISPOSITION_MISSING` finding visible through `review_memory(mode="adoption")` and `review_item_context` → approve via `apply_proposal` → reviewed-none disposition records the approver's `why` verbatim → page resurfaces in `review_memory(mode="relation-debt")`.
- All tests were written red-first (failing output captured before each implementation).

## Deferred follow-ups (review findings, out of scope here)

1. **Structured detail is lost when `semantic_writes.preflight_creation` raises**: `SemanticWriteError` carries only code+reason, so a hard-blocked proposal's findings reach the agent as one generic line. ~~Fixing this means attaching findings to the error in the semantic-writes layer.~~ **Largely resolved by #277 on main** (the raise now names the blocking findings in its message, which flows straight into this PR's `_contract_validation` detail; the code-prefix matching here is unaffected). Remaining nicety: structured (non-string) findings on the error object.
2. **Reconciliation-supersede proposals get no propose-time contract preview** despite routing through the same `_reviewed_creation` path at apply. Design deliberately scoped propose-time validation to compilation/supersession; extending it needs the reconciliation payload mapping done with the same exactly-as-`_route_apply` discipline.

https://claude.ai/code/session_018wa5EDTeP2xFFyzQA5BkAw


Artexis10:
**CI note — the py3.11 `tests` failure is a pre-existing environment-sensitive test, not this PR.**

`test_continuation_checkpoint.py::test_true_kill_after_pending_temp_fsync_is_cleaned_boundedly[retry]` failed 3× on the py3.11 runner (py3.13's first-attempt failure, a different test in the same file, greened on rerun). Root cause of the fragility: the orphan-cleanup sweep in `exomem_continuation_checkpoint.py` runs under `MAX_PRUNE_LOCK_SECONDS = 0.05` — a 50 ms monotonic deadline that silently aborts the enumerate+unlink sweep when exceeded, leaving the `.pending-*.tmp-*` file the test asserts is gone. A contended CI runner blows 50 ms easily; local runs never do (this test passes in two full-suite runs on this exact branch locally, and the continuation subsystem has zero commits between this branch's base and current main). Same flaky family was already observed on PR #234's CI and is queued for hardening — suggested fix there: inject a larger deadline in tests or loop the retry-cleanup like the \[prune\] variant already loops 20×.

PR #282, opened 2026-07-20
fix: cache the corpus context on a stat census and size the edge budget to measured write cost

Artexis10:
## What was measured (production corpus: 2,444 KB pages, laptop replica)

Before this branch, on the real vault:

| Stage | Cold | Warm |
|---|---|---|
| `note()` validate-only end-to-end | 40.2 s | 12.5-14.6 s |
| `build_corpus_context` (per call; up to 3x per one-shot write) | dominates cold | ~7.0 s |
| BGE model load (lazy, idle-evicted ~15 min in every mode) | 34.0 s | 0 |
| BM25 corpus build (lazy, suggest lane) | 40.0 s | 0 |

The HA edge abandoned every mutation-capable request at 15 s and deliberately
never replays, so nearly every write 502'd at the caller while the origin kept
committing: acknowledgement loss, while reads stayed fast.

After this branch, same vault, same machine:

| Stage | Result |
|---|---|
| `build_corpus_context` warm | **219.8 ms** (cache hit; census scan is 217.5 ms of it) |
| `note()` validate-only end-to-end warm | **2,005-2,067 ms** (3 consecutive runs) |

## What changed

1. **Permanent fix - corpus-context cache** (`semantic_contract.py`): the
   fully-built context is cached per vault root and served only when a
   stat-level census of every filesystem input matches exactly.
   `EXOMEM_DISABLE_CORPUS_CACHE=1` is the kill switch.
2. **Edge budget** (`deploy/cloudflare-ha`): default `MCP_TOOL_TIMEOUT_MS`
   15 s -> 60 s (under Cloudflare's ~100 s cap). **Live deploys that pin the
   old var must update it and redeploy the worker.**
3. **Acknowledgement-loss recovery** (`writer_lease.py`): implicit idempotency
   replay TTL 60 s -> 600 s; retrying a byte-identical write after an
   abandoned request replays the committed result (with the written slug)
   instead of failing against the existing page.
4. **Commit-stage semantics preserved** (`note.py`): an earlier iteration of
   this branch gated the whole post-commit corpus-aware pass behind
   `suggestions`; that was reverted on discovering
   `test_note_suggestions_knob.py` locks the dedupe sweep as a deliberate
   guardrail ("suggestions=False must skip the related-links pass and keep
   the sweep"). The sweep stays on in every mode; its cost is inside the
   widened budget and now visible per-write in the timing log.
5. **Write-path observability** (`note.py`, `semantic_contract.py`):
   content-free per-stage timings (preflight/commit/advisory/total and corpus
   census/build ms) logged on every write, so the next slow write is
   attributable from service logs in minutes, not hours.
6. **Classification** (`commands.py`): `remember` with `validate_only=true`
   is read-only (mirrors the existing `edit_memory` carve-out). Verified
   write-free by inspection: `note()` returns its preflight before any commit,
   path resolution uses `create_parents=False`, and all auxiliary work is
   planning-only. Draft commits remain writes.

## Why the invalidation is correct

The census records `(path, size, mtime_ns)` for **every filesystem input the
build reads** - the input set was enumerated by reading the full call tree:

- every `.md` under the KB, walked exactly like `_build_identity_census`
  (recursing everything, refusing symlinks/reparse points/non-regular files -
  the census returns "uncacheable" where the build would raise);
- every `.md` vault-wide, walked exactly like `vault.walk_vault_md` (same
  skip-dirs, same sync-conflict exclusion, same link-following semantics);
- `Knowledge Base/_access.yaml` (page eligibility via `access.access_tier`);
- `Knowledge Base/_Schema/relation-registry.yaml` and
  `semantic-language-registry.yaml` (presence AND absence are recorded);
- core registries ship in the package and are process-invariant.

Serving rule: current census == stored census. Storing rule: the census must
be **identical before and after** the build, so a context built from a tree
that changed mid-build is never stored. Caller-supplied registry objects must
be content-fingerprint-identical (`core_version`/`extension_hash`,
`schema_version`/`content_hash`) to a fresh disk load, or the cache is
bypassed - synthetic/proposal registries can never be served a disk-derived
context or poison it. Candidate-bearing builds always take the uncached path
(their inline candidate handling has subtly different unreadable-file
semantics than `with_candidate`, so it is not cached rather than approximated).

**Why not max-mtime:** Syncthing materializes remote edits with the source's
timestamp, which can be older than every local mtime - a max-mtime key would
silently serve a stale corpus. The per-path census catches this because the
synced file's `mtime_ns` (and normally size) differs from the entry recorded
for the previous content. Locked by a regression test that rewrites a page
with byte-identical length and an mtime one hour older, and asserts a rebuild.

`SemanticCorpusContext` is deep-frozen (`MappingProxyType`, frozensets,
tuples; the registry is defensively copied at construction), so a shared
cached instance cannot be mutated by consumers; `with_candidate` returns a
new context.

## Fixed permanently vs mitigated

- **Fixed permanently:** the O(corpus) re-parse on warm writes. Warm validates
  and both commit-time re-validations now cost a ~0.2 s census instead of
  ~7 s each; each committed write invalidates once and pays one rebuild
  afterwards. One-shot remember: 3 corpus parses -> 1.
- **Mitigated:** the rebuild itself is still O(corpus) (~7.8 s today) and
  runs once after every committed write and on every cold start; the census
  scan is O(corpus stats) (~0.2 s at 2.4 k pages). At sharply larger corpora
  the rebuild-after-write and the census will grow linearly - the next step,
  if ever needed, is incremental context updates keyed on the same census.
- **Mitigated:** cold model/BM25 loads (34 s / 40 s) are untouched product
  behavior (idle reaper runs in every mode); under the 60 s edge budget they
  no longer break writes. `EXOMEM_RELEASE_GPU_WHEN_IDLE=0` or
  `EXOMEM_MODE=performance` remove them at RAM cost.

## Remaining risk (explicit)

- The undetectable census case: new content at the same path with
  byte-identical size and identical 100 ns mtime. Not producible by the
  engine's own temp+rename writes nor by observed sync behavior; would
  require deliberate forgery or a filesystem without ns timestamps on the
  vault volume.
- One cached context per vault stays resident (bounded to 2 vault roots),
  raising steady-state RSS by roughly one corpus parse's worth of objects.
  Unmeasured; the kill switch restores the old memory profile.
- Read-only `remember validate_only` now runs without the mutation boundary,
  so it can observe a torn multi-file batch mid-commit exactly like every
  other read; any resulting inconsistency is caught by the fresh under-lock
  re-validation at commit. This matches the existing `edit_memory` posture.
- Full test suite on Windows excludes two pre-existing Unix-only modules
  (`test_hosted_restore_candidate.py` imports `fcntl`;
  `test_continuation_checkpoint.py` does `os.fsdecode(b"\xff")` at import) -
  both untouched by this branch and covered by Linux CI.
- 11 pre-existing Windows failures in `test_adopt.py` (semantic census
  scan-only, `markdown_files_scanned == 0`) and
  `test_adoption_proposals.py::test_apply_proposal_supersession_uses_replace_cas`
  were verified byte-identical on a pristine pre-branch checkout on this
  machine (via `git stash` round-trip): not introduced by this branch.
- Suite posture: the wider suite runs with the cache disabled via the same
  conftest convention used for embeddings/CLIP/watcher/warmup, because many
  tests monkeypatch corpus-affecting internals (access tiers, registries,
  walks) that no filesystem census can observe; the dedicated
  `test_corpus_context_cache.py` suite opts back in. Production inputs are
  filesystem-only, which is exactly what the census covers; cache behavior in
  production is additionally observable per call via the census/build log
  line.

## Deploy checklist (not done by this PR)

1. Merge, pull on the desktop (writer), restart the exomem service there
   (laptop follower next).
2. `npx wrangler deploy` the edge worker AND set the live
   `MCP_TOOL_TIMEOUT_MS` var to `60000` (the pinned old value overrides the
   new code default).

https://claude.ai/code/session_01FhLLSASsBqv7DNn7vRp1eJ

---

https://claude.ai/code/session_01FhLLSASsBqv7DNn7vRp1eJ

### DIFF
--- a/deploy/cloudflare-ha/src/worker.js
+++ b/deploy/cloudflare-ha/src/worker.js
@@ -314,7 +314,12 @@ async function proxyMutationRequest(request, env, lease) {
   request = withMutationRequestId(request);
   const requestId = request.headers.get("x-exomem-request-id");
   const shortTimeout = Number(env.ORIGIN_TIMEOUT_MS || 2500);
-  const toolTimeout = Number(env.MCP_TOOL_TIMEOUT_MS || 15000);
+  // 60s, not 15s: a governed write validates the draft against the full corpus
+  // and re-validates under the creation lock, measured at 12-45s warm on the
+  // 2026-07 production corpus (2.4k pages). Abandoning at 15s guaranteed the
+  // origin kept committing after the edge stopped waiting — acknowledgement
+  // loss on nearly every write. 60s stays under Cloudflare's ~100s proxy cap.
+  const toolTimeout = Number(env.MCP_TOOL_TIMEOUT_MS || 60000);
   const holder = lease.holder;
   let candidate = candidateForHolder(env, holder);
 
--- a/deploy/cloudflare-ha/wrangler.toml.example
+++ b/deploy/cloudflare-ha/wrangler.toml.example
@@ -11,8 +11,11 @@ LAPTOP_ORIGIN = "https://exomem-laptop.example.com"
 ORIGIN_TIMEOUT_MS = "2500"
 # MCP tools can legitimately take longer than a connectivity probe. Keep this
 # independent from the short connectivity probe; ambiguous tool calls are
-# never cross-replica replayed regardless of timeout.
-MCP_TOOL_TIMEOUT_MS = "15000"
+# never cross-replica replayed regardless of timeout. Governed writes validate
+# against the full corpus (measured 12-45s warm at 2.4k pages), so this must
+# comfortably exceed a warm write and stay under Cloudflare's ~100s proxy cap.
+# An abandoned mutation is acknowledgement loss: the origin commits anyway.
+MCP_TOOL_TIMEOUT_MS = "60000"
 # Behavioral compatibility, not exact package versions. During an incompatible
 # rolling deployment, expand to "1,2", roll every replica, then contract to "2".
 SUPPORTED_RUNTIME_CONTRACTS = "1"
--- a/src/exomem/commands.py
+++ b/src/exomem/commands.py
@@ -5205,6 +5205,13 @@ def invocation_is_read_only(command: Command, kwargs: dict[str, Any]) -> bool:
         return mode == "audit"
     if command.name == "edit_memory":
         return kwargs.get("validate_only") is True
+    if command.name == "remember":
+        # A validate-only remember builds and returns an immutable draft and
+        # writes nothing (note() returns its preflight before any commit), so
+        # it needs neither writer authority nor the mutation boundary. Any
+        # inconsistency from reading beside a concurrent write is caught by
+        # the fresh under-lock re-validation at commit time.
+        return kwargs.get("validate_only") is True
     return False
 
 
--- a/src/exomem/note.py
+++ b/src/exomem/note.py
@@ -32,6 +32,7 @@ import datetime as dt
 import logging
 import os
 import re
+import time
 from dataclasses import dataclass, field
 from pathlib import Path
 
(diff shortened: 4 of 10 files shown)


Artexis10:
Superseded by #285, which integrated this PR's corpus-context cache, stage timings, 60s edge budget, 600s replay TTL, and read-only validate_only wholesale (its body credits this PR directly) — and main has moved past it, leaving this branch CONFLICTING. Closing rather than resolving conflicts to re-land already-merged work. The deploy follow-through this PR documented still applies: wrangler deploy with the live MCP_TOOL_TIMEOUT_MS var updated to 60000, then desktop pull + service restart.

PR #286, opened 2026-07-21
fix: skip readonly subtrees in audit_fix instead of aborting the pass

Artexis10:
## The defect

`audit_fix` walks every compiled page without consulting `_access.yaml`. A `readonly` subtree refuses every write with **no override**, so a planned wikilink canonicalisation for one reaches the write layer, raises `WRITE_REFUSED`, and aborts the entire batch.

Found on a real vault today:

```
WRITE_REFUSED: Knowledge Base/Products/Call Center AI/Strategy.md:
path is in a `readonly` tree (_access.yaml): findable but write-protected
```

Three pages under a readonly `Products/` tree blocked **seventeen** perfectly writeable pages. `maintain_memory(mode="fix")` could repair nothing at all. Since `readonly` is the documented, supported configuration, any vault using it has a permanently unusable link fixer.

## Two defects, one cause

1. **The dry run promises work it cannot do.** Readonly pages were reported under `fixed`, so `maintain_memory(mode="fix", dry_run=true)` returned `fixed: 20` when a real run could apply none of them. A preview that disagrees with the real run isn't a preview.
2. **The real run fails closed on the first refusal**, taking every unrelated writeable page down with it.

## The change

The pass now consults `access.access_tier` per page — the same guard `audit.py` already applies at three other sites (lines 1447, 1699, 1863) — and skips anything not `TIER_READ_WRITE`.

The count surfaces as `skipped_readonly` rather than being dropped silently: a caller reconciling a dry run against the audit's finding count needs to account for the gap, and a silent skip would just relocate the confusion.

## Tests

Both confirmed to **fail against the unfixed pass** (verified by stashing `src/` and re-running), then pass with it:

- `test_readonly_subtree_does_not_block_fixing_writeable_pages` — a writeable page beside a readonly one is still canonicalised; the readonly page is left byte-identical; the skip is reported.
- `test_dry_run_does_not_promise_fixes_it_cannot_apply` — the preview never lists a readonly page under `fixed`.

## Verification

- 22 passed across `test_audit_fix.py` and `test_access.py`
- `ruff check` clean

https://claude.ai/code/session_012z8W4eQFrREL758npaVPH6

### DIFF
--- a/src/exomem/audit_fix.py
+++ b/src/exomem/audit_fix.py
@@ -50,9 +50,9 @@ import re
 from dataclasses import dataclass, field
 from pathlib import Path
 
+from . import access, indexes
 from . import audit as audit_module
 from . import find as find_module
-from . import indexes
 from .vault import (
     PlannedWrite,
     WikilinkResolver,
@@ -63,7 +63,6 @@ from .vault import (
     parse_frontmatter,
     render_wikilink_target,
 )
-
 
 log = logging.getLogger(__name__)
 
@@ -91,6 +90,10 @@ class AuditFixReport:
     files_rewritten: int = 0
     summary: dict[str, int] = field(default_factory=dict)
     dry_run: bool = False
+    # Pages passed over because they sit in a `readonly` subtree. Reported
+    # rather than silently dropped: a caller comparing a dry run against the
+    # audit's finding count must be able to account for the difference.
+    skipped_readonly: int = 0
 
     def as_dict(self) -> dict:
         return {
@@ -107,6 +110,7 @@ class AuditFixReport:
             "files_rewritten": self.files_rewritten,
             "summary": self.summary,
             "dry_run": self.dry_run,
+            "skipped_readonly": self.skipped_readonly,
         }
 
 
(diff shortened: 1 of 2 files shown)


Artexis10:
## Correction to the numbers in this PR and its commit message

Verified against production after deploying 0.27.1. The description above says "three readonly pages blocked seventeen perfectly writeable ones." **That ratio is wrong.** It was inferred from a five-item sample rather than counted.

Measured on the real vault with the fix live:

```
skipped_readonly: 77
fixed: 2
readonly paths listed under 'fixed': []
```

Of the 20 pages the old dry run reported as fixable, **18 were in readonly trees and 2 were writeable** — not 3 and 17. Across the full walk, 77 pages sit in readonly subtrees.

The defect, the diagnosis, and the fix are unaffected: a single readonly page aborted the entire batch, so the fixer was unusable on any vault using `_access.yaml`. That holds at any ratio. Only the illustrative counts were wrong, and they are wrong in the merged commit message too, which cannot be corrected without rewriting main.

Confirmed working end to end: the two writeable pages were canonicalised, and readonly pages were left byte-identical.

PR #318, opened 2026-07-24
docs: propose MCP call ledger (design-only OpenSpec change)

Artexis10:
Design-only OpenSpec change — no `src/` code. Proposes a per-call ledger to close the observability gaps a live investigation surfaced on 2026-07-24.

## Why (all verified against code/production)

1. **Trace history is evicted under load.** `CallTraceMiddleware` (`server.py:77-145`) does trace every call, but a read-only-replica traceback storm cut `exomem.log` retention to ~5 hours, evicting the tool-call history operators need. (Separately hotfixed in #317.)
2. **Refusals are logged as successes.** The tool wrapper (`command_surface.py:234-242`) returns an error envelope rather than raising, so the middleware records `event=tool_success`; the `OpError.code` is never durably captured.
3. **No durable per-call record.** The only durable per-request store is `.idempotency.sqlite` — mutations-only, keyed/overwritten, TTL-pruned.

## What

An append-only, hash-chained, redacted JSONL ledger, captured at two seams (`CallTraceMiddleware` for identity/timing/arg-shape + `command_surface` wrapper for outcome/`OpError.code`), written **outside the vault and outside the writer lease** so it records during the incidents it exists to explain. Arguments recorded as shape+hash, never values; caller identity from the already-computed `mcp_retry_scope()` principal hash. Monotonic sequence + per-row `prev_hash` chain make eviction detectable; content-addressed rotation modeled on `log.md`.

**Non-goals:** not fixing the writer fail-back; not a metrics/OTel system; not changing tool behavior.

## Validation

`openspec validate --changes --strict` → 72 passed / 0 failed; `--specs --strict` → 23 passed / 0 failed (CI's exact invocation).

## For human review — open questions (in `design.md`)

`target_paths` verbatim vs hashed (filenames can leak); argument-hash salt/oracle; whether guard-content rejections should be `refused` vs `error`; archive pruning policy; one ledger file vs split reads/writes; and the load-bearing assumption that all tool calls dispatch in the main process (a `writer_id` field is reserved as fallback). Committed as `docs:` so it triggers no release.

### DIFF
--- a/openspec/changes/add-mcp-call-ledger/.openspec.yaml
+++ b/openspec/changes/add-mcp-call-ledger/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-24
--- a/openspec/changes/add-mcp-call-ledger/design.md
+++ b/openspec/changes/add-mcp-call-ledger/design.md
@@ -1,0 +1,225 @@
+## Context
+
+Exomem already has a per-call tracer. `CallTraceMiddleware` (`src/exomem/server.py:77-145`) is
+installed on both profiles — hosted (`src/exomem/server.py:241`) and local
+(`src/exomem/server.py:255`). For each call it resolves a `request_id`
+(`mcp_request_id()`, a UUIDv4 that honors an inbound `x-exomem-request-id` header,
+`src/exomem/command_surface.py:287-315`), binds it through the synchronous wrapper with
+`mcp_request_context` (`src/exomem/server.py:90`), and emits `event=tool_start` before the call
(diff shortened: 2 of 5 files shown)


Artexis10:
Closing this design as stale against current main. Independent review found that its worker-thread ContextVar outcome bridge cannot propagate back to async middleware, the host-global ledger is only process-locked, the durability contract is internally inconsistent, and path/hash redaction decisions remain unresolved. #320 and #349 also changed the observability and receipt baseline. A future all-call ledger should start as a narrowed replacement spec against those current contracts.

PR #326, opened 2026-07-25
feat(governance): add the governance kernel — policy, compiler, membership, pure evaluator

Artexis10:
Implements OpenSpec change `add-governance-kernel` — wave 1 of the user-sovereign knowledge governance plan (#316). **Inspection-only: zero enforcement.** The kernel can load `_Governance/` policy, resolve scope membership, and answer "what is the effective disclosure ceiling for this item/audience/purpose" — but nothing in the retrieval or read path consults it until `add-release-gate`. An AST-level import scan confirms no module outside `src/exomem/governance/` touches the package, and the latency gate passes untouched.

## What's in the kernel

- **Policy source** (D1): strict YAML in `Knowledge Base/_Governance/` (`scopes/`, `rules/`, `grants/`), JSON-Schema-shaped validation with findings-not-exceptions; unknown *field* = fail-closed compile error, unknown *file* = ignored; `governance_version: 1`, ULID ids. The `.governance.sqlite` sidecar stores the compiled snapshot + meta only — rebuildable, never the enforcement authority, and `write_snapshot` refuses empty/blocked policies.
- **Fingerprinted live reload** (D2): `access.py`-style per-file stat signature + content hash; missing `_Governance/` → process-cached `EMPTY_POLICY` singleton and a guaranteed open fast path (no page parse, no sidecar open — pinned by test).
- **Fail-closed refusal** (D3, hardened in review): a `(conflicted copy` sibling — case-insensitive, any extension, including look-alike filenames — or any compile error refuses the compile. With a prior good compile, the last good policy stays in effect with findings attached. With **no** prior good state, the policy is **BLOCKED**: a sentinel distinct from `empty`, floored at L0 by `decide_paths`. A refused compile can never present as "no governance, fully open."
- **Query-time membership** (D4): selector evaluation over the in-hand `ParsedPage` (glob/project/tag/type/class/ref + excludes), bounded LRU memo keyed `(fingerprint, path, mtime_ns)` — no index-time table, no new `index_sync` component.
- **Pure evaluator** (D5): `min(org_cap, max(grants, min(standing_rules)))`, default OPEN L6, order-free (verified by exhaustive permutation), both purpose-semantics directions tested; `options`/`notice`/`bridge` merge pinned to lexical rule-id order so `add-release-gate` inherits a tested contract.

## Review provenance

Sonnet executor implementation (red-first per task group) + independent read-only review with fresh-process probes, one correction round, one targeted recheck:

- **HIGH (caught in review, fixed, recheck-verified)**: cold-start compile failure originally returned an `EMPTY_POLICY`-derived object — `.empty == True` — so a typo'd `ceiling:` or a sync conflict on first load silently meant *fully open* with findings inert. Now BLOCKED → L0 floor, findings surfaced, no snapshot persisted; fresh-process probes confirm both reproductions.
- **MEDIUM ×2 (fixed, recheck-verified)**: case-sensitive conflict marker (a capital-C conflicted copy was only caught by an accidental `duplicate_id` backstop); conflict detection racing the files-empty short-circuit (a dir of only conflicted files returned silent EMPTY).
- **Task 6.1 deviation (adjudicated)**: kernel read leaves live in an internal `KERNEL_LEAVES` registry rather than `commands.py` — `add-governance-tools` plans its own router and nothing downstream requires COMMANDS entries now; keeping user-facing surfaces at zero honors this change's scope.
- **Flagged forward, not a defect here**: `add-release-gate`'s planning docs never reference `.blocked`/`decide_paths`/`policy.load`, and its `annotate_hits` design suggests direct `membership`/`decide` calls — the gate's implementer must apply the `.blocked` fail-closed contract or the cold-start hole reappears at the point of real enforcement. This is baked into the release-gate lane brief. Also noted: a `_Governance/` dir containing only unrecognized files returns EMPTY without the design-promised warning (low, loader nit, follow-up).

## Verification

`56 passed` across the five governance suites + scaffold no-leak; latency gate green ×3 independent runs (find path untouched); scoped ruff clean; `openspec validate add-governance-kernel --strict` valid. Fresh-process probe outputs and the full review/recheck records are preserved in the session evidence trail.

https://claude.ai/code/session_01Cz18XCnax9FiZnkaXTn7sj


Artexis10:
Note for whoever merges this: the `purpose` semantics shipped here allow a *widening* path — a rule with `purpose_condition: outside` can be escaped by a client declaring that purpose, since `purpose` arrives verbatim from the caller and is never attested. That is inert today (nothing calls the evaluator; this change is inspection-only), but it should not be built on.

The hardening lands in the stacked `add-release-gate` PR as an owner-approved invariant: **purpose may only narrow, never widen** — `decide(purpose=X).level <= decide(purpose=None).level` for every X, implemented by evaluating the lattice twice and taking the lower ceiling, with a property test. Identity/audience-conditioned grants still raise ceilings normally; only self-declared widening is removed.

If this merges before the gate, the invariant arrives with the gate. If you want them independent, the `decisions.py` hunk can be cherry-picked back here first.

PR #509, opened 2026-08-15
feat(hosted): add the reviewer bootstrap driver and promotion-evidence harness

Artexis10:
Encodes the virgin-install reviewer OAuth bootstrap and the cohort-promotion
evidence path, both of which were previously hand-run curl against a hard
30-minute window.

## reviewer_bootstrap.py — three traps fixed, each found by running it

- `urlopen` followed the 303 from `/oauth/authorize`, so the caller saw the
  final 200 page and the `exomem_oauth_tx` cookie set on the redirect was
  dropped. Now surfaces 3xx via a no-redirect opener.
- The sibling loop only covered Claude. `promote-cohort` requires both a
  `claudeArtifactId` and an `openaiArtifactId`, and each canary credential is
  welded to the bootstrap's own assignment and generation, so a later pass
  cannot supply the missing one.
- `register_pinned` omitted `registeredAppIdSha256`. The stage is matched with
  `IS NOT DISTINCT FROM`, so an OpenAI client whose stage carries the digest
  matched no stage row and returned a bare `400 INVALID_REQUEST`.

## promotion_evidence.py — new

`observe` / `sign` / `import` / `promote`. It deliberately does not invent the
facts it signs: the seven operation flags are read from a results file
describing a real client session, and the six counts are queried from the
control plane with the run aborting unless each is exactly 1. `--rehearse`
exercises every query and gate and then refuses to emit, so a dry run cannot
manufacture evidence for a session that never happened.

Verified against a live tenant: counts resolve, both platform shapes match the
contract (41 fields Claude / 42 OpenAI), and the promote path resolves the
compare-and-swap digest correctly.

## Why the timing matters

The bootstrap authority is capped server-side at 30 minutes, the rollout
assignment inherits that expiry, and `storeClientArtifact` additionally
requires the assignment still be active — so the evidence run for both
platforms, signing, both imports and promotion all land inside one window.
That constraint applies to the first promotion only: afterwards a live cohort
admits a real client directly and an ordinary assignment may live 7 days.

### DIFF
--- a/scripts/promotion_evidence.py
+++ b/scripts/promotion_evidence.py
@@ -1,0 +1,514 @@
+#!/usr/bin/env python3
+"""Observe, sign and import promotion evidence for a hosted cohort candidate.
+
+`scripts/sign-directory-evidence.py` is a different tool: it signs marketplace
+*directory* evidence. This one produces the record `validatePromotionEvidence`
+accepts, which is what `import-artifact` and then `promote-cohort` consume.
+
+Like that script, this deliberately does not invent the facts it signs. The
+seven operation flags describe a real client session and are read from a results
+file the operator writes after running it; the six counts are queried from the
+control plane and the run aborts unless each is exactly 1. A harness that
+defaulted those to true/1 would turn the signed-evidence chain into decoration.
+
+Everything here is timing-critical for the FIRST promotion only. The bootstrap
+authority is capped server-side at 30 minutes, the rollout assignment inherits
+that expiry, and `storeClientArtifact` requires the assignment still be active --
+so observe, sign, both imports and promote all have to land inside it. Rehearse
+against an existing tenant before opening a window.
+
+    observe  --platform claude --results results-claude.json
+    sign     --platform claude
+    import   --platform claude
+    promote
+
+Subsequent promotions do not need this urgency: once a cohort is live a real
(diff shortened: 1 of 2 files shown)


Artexis10:
Closing as superseded — this landed through #595 (`feat(hosted): rescue the promotion harness and fix the OpenAI sibling identity`) and the work that followed it.

Both files this PR adds already exist on `main`, and `main`'s versions are strictly ahead:

| file | lines on `main` that this PR lacks |
|---|---:|
| `scripts/promotion_evidence.py` | 59 |
| `scripts/reviewer_bootstrap.py` | 146 |

Merging it would **delete 205 lines of subsequent work**.

Everything this branch has that `main` does not is an older form of something `main` already does better. The two that matter:

- `--repo` defaulted to a hardcoded home directory containing a username. That is the `absolute_local_path` finding that has kept `package build` red here since 2026-08-15. `main` derives it: `Path(__file__).resolve().parent.parent`.
- the identity count interpolated the owner straight into SQL — `select count(*) from users where id='{owner}'`. `main` does not.

Confirmed no capability is lost: `identity_count`, `paired_run_hmac_sha256`, `--stage-id`, `routableCellCount`, `LOOPBACK_REDIRECT` and the both-artifacts promotion guard are all present on `main`.

Reopen if I've misread something — nothing here is deleted, the branch is intact.

PR #587, opened 2026-08-16
fix(ops): fail the upgrade when the installed version does not change

### LINKED ISSUE #578: scripts/upgrade.ps1 reports a successful install while leaving the old version in place
`scripts/upgrade.ps1` reports that it is installing, exits its install step without error, and leaves the previous version in place. The operator sees a normal-looking run and a service still on the old build.

Hit while deploying 0.52.3 onto the live personal cell. Caught only because the version was checked independently afterwards — the script's own output gives no indication anything went wrong.

## Reproduction

```
$ pwsh -NoProfile -ExecutionPolicy Bypass -File scripts/upgrade.ps1
Service 'exomem'
  venv:      C:\Users\hugoa\Desktop\projects\exomem-service-ha\.venv\Scripts\python.exe
  installed: 0.52.2
  repo:      0.52.2
Installing exomem[embeddings,media] into ...\.venv\Scripts\python.exe...
   ... proceeds to doctor preflight ...

$ ...\.venv\Scripts\python.exe -c "import importlib.metadata as m; print(m.version('exomem'))"
0.52.2
```

0.52.3 was published to PyPI and resolvable at the time (`pip index` / direct install both fine). Installing manually with the same interpreter worked immediately:

```
$ uv pip install --python ...\.venv\Scripts\python.exe --upgrade "exomem[embeddings,media]==0.52.3"
Resolved 135 packages in 839ms
Prepared 1 package in 734ms
Uninstalled 1 package in 183ms
Installed 1 package in 429ms
 - exomem==0.52.2
 + exomem==0.52.3
```

So the package, the index, the interpreter and `uv` were all fine. The failure is in the script path.

## Two distinct defects

**1. The install step does not fail loudly when it does not install.** `Install-ExomemPackage` runs, the script prints `Installing ...`, and execution continues to `Repair-TorchCuda` and the doctor preflight as though it had succeeded. `Invoke-LoggedNative` (`scripts/_service-common.ps1:159-169`) returns `$LASTEXITCODE`, but nothing between the install and the preflight asserts that the installed version actually changed. The script already computes `$before` and `$after` and prints `Installed version: $before -> $after` — that line is a report, not a check. When `$after -eq $before` and a different version was requested, the run should stop.

**2. Under Windows PowerShell 5.1 it aborts on `uv`'s stderr banner.** `Invoke-LoggedNative` merges native stderr with `2>&1`, so `uv`'s ordinary informational line becomes an ErrorRecord:

```
uv.exe : Using Python 3.13.11 environment at: ...\.venv
At ...\scripts\_service-common.ps1:166 char:12
    + CategoryInfo          : NotSpecified: (Using Python 3....ervice-ha\.venv:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

That is not an error condition — `uv` writes progress to stderr by design. Under `pwsh` (7.x) the script gets further and reaches the preflight, but the install still does not take, so defect 1 is independent of the shell and is the more serious of the two.

## Why this is worth fixing rather than documenting

A deploy script whose failure mode is "prints success, changes nothing" is worse than one that errors: the operator proceeds to verification, the service restarts cleanly on the *old* build, and every subsequent observation is attributed to a release that was never deployed. In this case the release contained the fixes being verified, so a silent no-op would have produced a confident and completely wrong conclusion about whether they worked.

The script is also the documented path — `CLAUDE.md` and the deployment docs say to use `scripts/upgrade.ps1` and specifically *not* plain `pip`, because the service runs a separate PyPI venv. So there is no sanctioned fallback an operator is expected to reach for.

## Asks

1. **Fail the run when the installed version does not change to the requested one.** `$before`/`$after` are already computed; turn the report into an assertion. If no explicit `-PackageVersion` was given, compare against the resolved target rather than skipping the check.
2. **Propagate the install's real exit status.** Establish whether `uv` returned nonzero and the code was dropped, or returned zero having done nothing. Both are fixable but they are different bugs, and the current logging cannot distinguish them.
3. **Stop treating native stderr as failure.** Capture stdout and stderr separately, or set `$ErrorActionPreference`/`2>$null` around the native call so an informational banner cannot abort the run under PowerShell 5.1.
4. **State the required shell.** The header comment says `pwsh -File scripts/upgrade.ps1`; if 7.x is genuinely required, refuse to run under 5.1 with a clear message instead of failing obscurely partway.
5. **Print the resolved target version before installing**, so `installed: / repo: / target:` makes a no-op visible at a glance.

## Also observed on the same run, filed here only to keep it together

The doctor preflight correctly **blocked** the restart on `rebuild_temp.orphans` (78 lexical rebuild temporaries, 6487 MB). That gate worked exactly as intended and is not a bug — but note the ordering consequence: the reaper that clears those temporaries shipped *in the release being deployed*, so the gate blocks the deploy that fixes the condition it is blocking on. Clearing it required an out-of-process `exomem maintain --reconcile` with the service stopped. Worth considering whether the orphan gate should be advisory when the incoming version is newer than the installed one, or whether the reaper should be reachable without a full reconcile.

Artexis10:
Closes #578

`scripts/upgrade.ps1` printed `Installing exomem[embeddings,media] into ...`, continued to the doctor preflight, and left the venv on the previous release. The service then restarted cleanly on the old build and every later observation was attributed to a release that was never deployed — including the verification of the fixes that release contained.

## What the evidence says about ask 2

The issue asks whether `uv` returned nonzero and the code was dropped, or returned zero having done nothing. **It returned zero.** Reproduced with a fake native command under both shells:

| shell | uv exit | `Invoke-LoggedNative` result |
|---|---|---|
| Windows PowerShell 5.1 | 0 | **throws** `NativeCommandError` / `RemoteException` on the stderr banner |
| pwsh 7.6.4 | 0 | returns `0` |
| pwsh 7.6.4 | 1 | returns `1` |

`$PSNativeCommandUseErrorActionPreference` is `False` by default in 7.6, so a nonzero exit does not throw: `Invoke-LoggedNative` returns it, `Install-ExomemPackage` throws `uv pip install failed`, and `$ErrorActionPreference = "Stop"` aborts the script. The failed run *reached the doctor preflight*, so that path did not happen — uv exited 0 without applying the change. The exit code was never dropped; there was no failure to propagate.

*Why* uv no-oped is not recoverable from the artifacts (the run's uv output was not retained). The most likely cause is uv's cached package index: the unpinned `--upgrade` resolved back to 0.52.2, while the manual `==0.52.3` forced a fresh lookup for that exact version and worked immediately. That is a hypothesis, not a finding — so the fix both defends against it (`--refresh-package exomem`) and makes the next occurrence self-diagnosing (the resolved target is printed, and a mismatch names both versions).

## Changes

**1. The report becomes a gate.** `Install-ExomemPackage` now resolves the concrete target with `uv pip install --dry-run --upgrade --refresh-package exomem` before installing, prints it as `target:` under `installed:`/`repo:`, and calls `Assert-ExomemInstallApplied` afterwards. It fails when nothing is importable, and when the interpreter does not end up on the resolved target. An explicit `-PackageVersion` stays assertable even if the resolve itself fails; if the target cannot be resolved at all, the run says so out loud instead of silently skipping the check. `before == after == target` (an already-current box) is correctly *not* a failure.

`--refresh-package exomem` is load-bearing rather than belt-and-braces: a target read through the same stale cache would agree with the stale install and vouch for it.

**2. Native stderr stops being a failure.** `uv` writes its entire plan to stderr and nothing to stdout, so the `2>&1` merge is necessary for any logging at all — the preference is what was wrong. `Invoke-ExomemNative` sets `$ErrorActionPreference = "Continue"` and `$PSNativeCommandUseErrorActionPreference = $false` **function-scoped**, so callers keep `Stop` semantics everywhere else. It also refuses up front when the executable is not on PATH, since `$LASTEXITCODE` would otherwise be left at the previous command's value and read as success.

**3. Ask 4 — the shell gate.** `upgrade.ps1` refuses 5.1 with a message naming the shell and the fix. 7.x is a genuine requirement, not a preference: `Invoke-WebRequest -SkipHttpErrorCheck` is 7.0+ only.

**4. `upgrade.sh`** — checked, not assumed. It never had defect 2 (bash does not treat stderr as failure, and `set -euo pipefail` already stopped a nonzero uv) but carried defect 1 verbatim: it computed `BEFORE`/`AFTER`, printed them, and checked nothing. Same resolve → show → assert sequence added via `_service-common.sh`.

**Skipped:** nothing from the five asks. Ask 2 is answered by evidence rather than by a code change, since the exit code was already propagated correctly.

## Red-then-green

Both defects were run against unmodified `origin/main` in a disposable worktree, using the same test file: **16 failed, 1 passed**. The one pass is `test_uv_informational_stderr_never_aborts_the_run[pwsh]` — defect 2 does not reproduce under pwsh 7, which is exactly why the test is parametrised over both shells.

Defect 1, `[pwsh]`, on `origin/main`:

```
>       assert result.returncode != 0, (result.stdout, result.stderr)
E       AssertionError: ('Installing exomem[embeddings,media] into ...
E         Resolved 135 packages in 839ms
E         Audited 135 packages in 12ms
E         DRIVER-COMPLETED
E         ', '')
E       assert 0 != 0
```

Defect 2, `[windows-powershell-5.1]`, on `origin/main`:

```
>       assert result.returncode == 0, (result.stdout, result.stderr)
E       AssertionError: (... (Using Python 3.... at: fake-venv :String) [], RemoteException
E             + FullyQualifiedErrorId : NativeCommandError
E          ')
E       assert 1 == 0
```

Green on this branch: `23 passed`.

## Coverage honesty

Defect 2 is shell-specific and **cannot** be covered by Linux CI: pwsh 7 does not raise on merged native stderr, so no test running under pwsh can be red for it. The `windows-powershell-5.1` lane skips wherever `powershell.exe` is absent, i.e. all of CI. It is red-then-green **locally on Windows only**, quoted above. Defect 1 and the unix parity run everywhere.

Nothing in the suite touches a real venv, index, or service — `uv` and the service interpreter are both faked, in `.cmd` and shebang form so the PowerShell and bash lanes each resolve the fake rather than the real `uv` on PATH.

## Consumer sweep

`Invoke-LoggedNative` is shared infrastructure. Every caller enumerated with `rg`:

- `scripts/install-service.ps1:176` (`uv venv`) — unchanged contract; also stops dying on `uv venv`'s own banner under 5.1.
- `scripts/_service-common.ps1` `Repair-TorchCuda` — unchanged contract.
- `scripts/_service-common.ps1` `Sync-ExomemUvCli` — unchanged contract.

It keeps returning a bare `Int32`, so all three are untouched. `Install-ExomemPackage` deliberately **does not** return the resolved version: it is called bare inside `Install-ReleaseVenv`, whose return value is assigned to `$python`, and an extra object on the output stream would have turned that into an array and broken the installer.

`install-service.ps1` shares `Install-ExomemPackage`, so a fresh venv that ends up empty now fails there too — which is the same defect class, and desirable.

Tests that pin the script text and had to move with it: `tests/test_container_distribution.py`, `tests/test_service_installers.py` (both pinned the exact `uv pip install` argument list).

## Not fixed here

The issue's closing observation — the doctor orphan gate blocking the deploy that ships the reaper for the condition it gates on — is a real ordering problem and deliberately out of scope for this PR. Worth its own issue.

`scripts/install-service.ps1` has the same latent 7.0+ requirement (`-SkipHttpErrorCheck`) and no shell gate. Left alone to keep this diff scoped to the reported script; a one-line `Assert-ExomemPowerShell7` call would close it.

## Local verification

`23 passed` for `tests/test_upgrade_guard.py`; `tests/test_container_distribution.py` and `tests/test_install_parity.py` green. Eight pre-existing failures in `tests/test_service_installers.py` are the documented native-Windows bash-interop breakage (`/bin/bash: C:UsershugoaDesktop...: No such file or directory`) — `install-service.sh` does not source `_service-common.sh`, so they are untouched by this change. Linux CI is the authority.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -327,10 +327,16 @@ pwsh -File scripts/upgrade.ps1
 # Explicit policy: -CliSync auto|always|never (default: auto)
 ```
 
+Requires PowerShell 7+; Windows PowerShell 5.1 is refused up front rather than
+allowed to fail obscurely partway.
+
 The service runs a PyPI-backed venv that is **not** the repo checkout, so `git pull`
-never touches it. `upgrade.ps1` locates that venv from the NSSM registry, upgrades
-it, repairs the CUDA torch build, gates on `doctor`, restarts, and then asserts the
-**live** `/health` version matches what it just installed. It then records a
+never touches it. `upgrade.ps1` locates that venv from the NSSM registry, prints the
+`target:` version it resolved before touching anything, upgrades it, **fails if the
+venv did not actually end up on that target** (a `uv` that exits 0 having applied
+nothing is the failure this catches — #578), repairs the CUDA torch build, gates on
+`doctor`, restarts, and then asserts the **live** `/health` version matches what it
+just installed. It then records a
 non-secret managed-install manifest and, in the default `auto` mode, aligns an
 existing uv-managed `exomem`/`kb` command to that exact verified release. `auto`
 never installs a command that was absent; `always` may install it; `never` leaves
--- a/scripts/_service-common.ps1
+++ b/scripts/_service-common.ps1
@@ -156,6 +156,69 @@ function Get-ExomemServiceEndpoint {
     return $result
 }
 
+function Assert-ExomemPowerShell7 {
+    <#
+    .SYNOPSIS
+      Refuse to run under Windows PowerShell 5.1, which cannot execute these scripts.
+    .DESCRIPTION
+      Not a style preference: the deploy path calls
+      `Invoke-WebRequest -SkipHttpErrorCheck`, which is PowerShell 7.0+ only. Under
+      5.1 the run used to die partway with a NativeCommandError on an ordinary `uv`
+      banner -- a message that named neither the shell nor the real requirement.
+    #>
+    param([string]$ScriptName)
+
+    if ($PSVersionTable.PSVersion.Major -ge 7) { return }
+    throw "$ScriptName requires PowerShell 7+ (pwsh); this is Windows PowerShell $($PSVersionTable.PSVersion). Re-run with: pwsh -File scripts/$ScriptName"
+}
+
+function Invoke-ExomemNative {
+    <#
+    .SYNOPSIS
+      Run a native command, return its exit code and captured output, and never let
+      its stderr masquerade as a failure.
+    .DESCRIPTION
+      `uv` writes its ENTIRE plan to stderr by design -- the "Using Python 3.13.11
+      environment at: ..." banner, "Resolved N packages", and the "+ exomem==X"
+      lines all arrive there and nothing arrives on stdout. So the streams have to
+      be merged with 2>&1 for the output to be logged at all.
+
+      The hazard is what that merge does under `$ErrorActionPreference = "Stop"`,
+      which every calling script sets: Windows PowerShell 5.1 raises a terminating
+      NativeCommandError on the FIRST merged stderr record, even when the command
+      exits 0. That turned an informational banner into an aborted deploy (#578).
+
+      Both preference assignments below are function-scoped -- PowerShell shadows
+      the caller's value for this frame only and restores it on return -- so the
+      calling script keeps "Stop" semantics everywhere else. $ErrorActionPreference
+      keeps a stderr line from being fatal; $PSNativeCommandUseErrorActionPreference
+      (7.3+; a harmless unused variable on 5.1) keeps a NONZERO EXIT from throwing
+      before the code can be returned, which is what leaves "uv failed" reportable
+      as something other than "uv did nothing".
+    #>
+    param(
+        [string[]]$CommandArgs,
+        [switch]$Quiet
+    )
+
+    $ErrorActionPreference = "Continue"
+    $PSNativeCommandUseErrorActionPreference = $false
+
+    # Without this, a missing executable leaves $LASTEXITCODE at whatever the
+    # previous command set -- readable as success.
+    if (-not (Get-Command $CommandArgs[0] -ErrorAction SilentlyContinue)) {
+        throw "Required command '$($CommandArgs[0])' was not found on PATH."
+    }
+
+    $merged = & $CommandArgs[0] @($CommandArgs[1..($CommandArgs.Count - 1)]) 2>&1
+    $code = $LASTEXITCODE
+    $lines = @(foreach ($item in $merged) {
+        if ($item -is [System.Management.Automation.ErrorRecord]) { $item.ToString() } else { [string]$item }
+    })
+    if (-not $Quiet) { foreach ($line in $lines) { Write-Host $line } }
+    return @{ ExitCode = $code; Lines = $lines }
+}
+
 function Invoke-LoggedNative {
     <#
     .SYNOPSIS
@@ -163,9 +226,7 @@ function Invoke-LoggedNative {
     #>
     param([string[]]$CommandArgs)
 
-    $out = & $CommandArgs[0] @($CommandArgs[1..($CommandArgs.Count - 1)]) 2>&1
-    foreach ($line in $out) { Write-Host $line }
-    return $LASTEXITCODE
+    return (Invoke-ExomemNative -CommandArgs $CommandArgs).ExitCode
 }
 
 function Get-ExomemPackageSpec {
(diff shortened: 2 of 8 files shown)


Artexis10:
CI green on 31956556125 (all 8 test shards, lint, E2E, latency, golden gate, `required CI gate`).

Evidence that the new lanes actually ran on Linux rather than skipping: the only skips reported from `tests/test_upgrade_guard.py` are the `windows-powershell-5.1` parametrisations ("Windows PowerShell 5.1 is not available"). No `pwsh is not available` and no `bash is not available` skips, so the pwsh lane and the unix parity lane both executed.

One CI-only failure was fixed along the way and is worth recording: `test_nothing_installed_at_all_is_fatal[pwsh]` passed on a wide Windows console and failed on the runner, because PowerShell's error formatter hard-wraps to the host width and injects ANSI — it had split `no exomem version is importable` across lines. The driver now reports the failure itself via `Write-Host` and the assertions run through a flatten helper, so they no longer depend on terminal width.

Issue #566, opened 2026-08-16
POLLY cell re-pays a doomed full graph rebuild every ~62 min: non-stabilization never arms the publication-refusal memo


Artexis10:
Observed in production on the POLLY cell (port 8766, uv-tool install at `%APPDATA%\uv\tools\exomem`). It is burning ~4.5 CPU cores continuously and has been doing so for at least two days.

## Symptom

`C:\Users\hugoa\ExomemPolly\logs\exomem.log` contains **34** occurrences of:

```
file watcher: pending epoch graph recovery failed
RuntimeError: epistemic graph rebuild did not stabilize after 2 attempts
```

spanning `2026-08-14T20:12:28Z` → `2026-08-16T08:55:53Z`, roughly every ~62 minutes. Each one is a full epistemic-graph rebuild that runs to exhaustion and is thrown away.

It **survives restarts and a version upgrade** — the earliest tracebacks name `epistemic_graph.py:1130` / `file_watcher.py:568`, the latest name `epistemic_graph.py:1731` / `file_watcher.py:603`. Same failure, different line numbers.

### 1. The retry-suppression memo never arms, so the doomed rebuild is re-paid every cycle

This is the one that explains the CPU.

The codebase already has the exact mechanism for this: `note_publication_refusal` / `publication_refusal_active` (`epistemic_graph.py:332-352`), which `file_watcher.py:597` checks before calling `rebuild_all()`, with the comment

> The same publication was refused recently; re-paying a full rebuild every 300 s is exactly what the contract's R2 forbids.

And `_rebuild_all_off_boundary`'s own attempt-exhaustion path calls it (`epistemic_graph.py:1391`), with the comment

> only the memo is new, so the next cycle does not re-pay the same doomed rebuild.

But `_rebuild_all_locked`'s non-stabilization path (`epistemic_graph.py:1728-1731`) raises **without** setting the memo:

```python
message = "epistemic graph rebuild did not stabilize after 2 attempts"
if projection_moved:
    raise GraphProjectionMoved(message)
raise RuntimeError(message)
```

That exception propagates *out through* `_rebuild_all_off_boundary` at line 1309 — **around** its `note_publication_refusal` at line 1391, which is on the fall-through path, not the exception path. The traceback proves this directly: the watcher receives `RuntimeError`, not the `GraphRebuildRegistrationError("GRAPH_SYNC_STABILIZATION_EXHAUSTED", …)` that line 1392 would have raised. So line 1391 never ran.

[...]

PR #588, opened 2026-08-16
fix(graph): bound the interactive join so a write never parks on a rebuild

### LINKED ISSUE #576: Interactive writes block unbounded on a full-corpus graph rebuild after the commit (75-155s vs 0.2s when not joined)
Interactive knowledge-base writes on the primary personal cell are taking **75–155 seconds** against a budgeted commit median of ~750 ms. The write itself is not slow. Substantially all of the latency is post-commit derived work that the caller is forced to wait for, and the dominant term is a full lexical corpus rebuild that runs inside the request and then aborts.

## The split

Six consecutive committed mutations, `mutations.jsonl` + `exomem.log`, single cell, 2026-08-16:

```
tool          ts_utc     total     to files-committed   post-commit tail
remember      10:40:50    93.2s          4.9s                88.3s
edit_memory   10:52:45   141.3s         38.2s               103.1s
edit_memory   10:59:41   138.6s          4.8s               133.8s
edit_memory   11:14:24   112.8s          4.3s               108.4s
remember      11:21:39   106.1s          3.9s               102.2s
edit_memory   11:23:46    75.8s          2.8s                73.1s
```

**94–96% of the user-visible wait happens after `canonical_files_committed`.** The user's data is durable at ~3 s and they wait another ~100 s.

The vault lock is not the constraint. Every one of these records `boundary_wait_ms` ≈ 10 ms and `boundary_hold_ms` 3–5 s:

```
09:15:02  130570ms  wait=7.75   hold=4304.79  remember     committed
09:26:27   66516ms  wait=9.10   hold=4098.44  remember     committed
10:52:45  141126ms  wait=8.20   hold=4894.23  edit_memory  committed
11:14:24  112615ms  wait=10.36  hold=5422.20  edit_memory  committed
11:23:46   75748ms  wait=57.69  hold=3216.41  edit_memory  committed
```

A 5-second hold cannot explain a 141-second call. This rules out lock contention — including #571's doomed second rebuild pass — as the explanation for interactive write latency.

## Where the tail goes

Full log window for the 11:23:46 `edit_memory`, from files-committed to return:

```
11:22:33.477  exomem.calls        phase=canonical_files_committed  affected_count=5
11:22:36.001  exomem.semantic_…   semantic corpus context event-hit pages=3037          +2.5s
11:22:36.012  exomem.file_watcher semantic posthoc batch
11:22:36.254  exomem.lexstore     lexical sidecar upsert deferred (VAULT_LOCK_NESTED:
                                  nested vault creation locks are forbidden);
                                  heals on next sync
11:22:36.460  exomem.mutation_lock mutation_lock_released
11:22:37.609  exomem.mutation_lock mutation_lock_acquired                                +1.1s
11:22:42.859  exomem.calls        phase=canonically_committed                            +5.0s
11:23:21.011  exomem.lexstore     lexical atomic publish aborted:
                                  live catalog changed during build                     +38.2s
11:23:46.401  exomem.mutation_lock mutation_lock_acquired                               +25.4s
11:23:46.455  exomem.calls        phase=returned
```

The chain:

1. The **incremental** lexical sidecar upsert is refused: `VAULT_LOCK_NESTED: nested vault creation locks are forbidden`. It attempts a vault lock while the mutation already holds one, so the cheap path is structurally unreachable from inside a mutation.
2. The refusal falls back to a **full atomic publish** — a rebuild of the whole lexical corpus (3037 pages) — executed synchronously inside the request.
3. That rebuild runs ~38 s and then **aborts**: `live catalog changed during build`. The work is discarded.
4. A further ~25 s elapses with no log line before the call returns. Unattributed; needs instrumentation.

So the caller pays a full corpus rebuild, gets nothing for it, and the sidecar still "heals on next sync".

## Scale

Across log rotations `exomem.log.5` → `exomem.log`:

```
208  VAULT_LOCK_NESTED                 first seen 2026-08-15T07:05:20Z
187  lexical atomic publish aborted    first seen 2026-08-15T01:59:26Z
106  VAULT_LOCK_TIMEOUT                first seen 2026-08-15T01:58:50Z
138  epistemic graph rebuild did not stabilize
```

208 nested refusals against 187 aborted publishes is close to 1:1 — essentially every write that reaches the lexical stage takes the doomed path.

## An unexplained magnitude change

The defect is present across the whole window, but the cost is not constant. Median duration of committed mutations, per hour:

```
2026-08-15T09    39 writes    median   10647 ms
2026-08-15T13    22 writes    median    9574 ms
2026-08-15T16    11 writes    median    4228 ms
2026-08-15T20    36 writes    median    6361 ms
--- service restart 2026-08-16T00:19:01Z ---
2026-08-16T01    14 writes    median   86355 ms
2026-08-16T08     2 writes    median  163022 ms
2026-08-16T09     6 writes    median  106395 ms
2026-08-16T11     3 writes    median  105996 ms
```

`VAULT_LOCK_NESTED` fires at a similar per-write rate in both windows (Aug 15T15: 17 writes / 15 refusals, median 8.2 s; Aug 16T09: 6 writes / 6 refusals, median 106 s). So the *frequency* of the fallback did not change — the *cost of one rebuild* grew roughly 10×. What changed is not established. Candidates worth checking before assuming corpus growth:

- accumulated lexical orphan temporaries slowing the rebuild's own directory work
- something in the 0.52.2 deploy that started at `00:19:01Z`
- a genuine corpus/sidecar size change

## Likely relationship to the orphan backlog

187 aborted atomic publishes plausibly explains the ~6.5 GB of abandoned `.lexical.sqlite.rebuild-<uuid>.tmp` files that #551 was filed for — an abort is exactly the event that strands one. If so, #563's reaper treats the symptom and this issue is the source, and the two feed each other: more orphans → slower rebuild → longer window for the catalog to change → more aborts.

**Not verified.** Stated as a lead, not a finding.

## Why this is the priority

A user issuing `remember` or `edit_memory` waits ~100 s for a write that was durable at 3 s. Beyond the wait, it produces the abandoned-write mode already documented: clients time out while the origin commits anyway, so the caller cannot tell success from failure and may reasonably retry a write that already landed.

## Asks

1. **Do not make the caller wait for derived-index work.** The commit is durable at `canonical_files_committed`; lexical publication is explicitly self-healing (`heals on next sync`). Returning after the canonical commit and letting the sidecar heal asynchronously would cut ~95% of the observed latency without changing durability.
2. **Make the incremental upsert reachable from inside a mutation.** `VAULT_LOCK_NESTED` means the cheap path can never run where it is most needed, so the expensive fallback is not a fallback — it is the only path. Whichever way this is resolved, the incremental path should be exercised by a test that runs *inside* a held mutation lock, since that is the condition under which it fails.
3. **Do not run a full corpus rebuild synchronously in a request at all**, independent of (2). A rebuild whose abort condition is "the live catalog changed during build" cannot be made reliable on a write path that is itself changing the catalog.
4. **Instrument the unattributed ~25 s tail** between the aborted publish and the return. It is a quarter of the latency and nothing logs it.
5. **Explain the ~10× magnitude change** across the `00:19:01Z` restart before assuming corpus growth.

Related: #571 (separate defect — ruled out as the cause here), #551 / #563 (orphan temporaries, plausibly downstream of this), #526 (`VAULT_LOCK_TIMEOUT` deferrals on the same path).

Artexis10:
Closes #576.

A write parks on a derived-graph rebuild that may never converge. Observed in production: a single-unit `observe_memory` append blocked **300 seconds**, while the full `remember` two minutes earlier took 14 s. The cost is not in the writing.

## Two unbounded join sites, not one

The original analysis found `writer_lease.py` `after_operation_guard` and reasoned that `mutation_guard` was "not an interactive path." That was wrong, and production proved it — `observe_memory` exits through exactly that path:

```
WARNING exomem.writer_lease: direct mutation graph join failed after canonical release
  File "...\writer_lease.py", line 2266, in mutation_guard
    graph_sync.wait_for_registered(root, state_root=self.config.state_dir)
exomem.graph_sync.GraphRebuildRegistrationError: GRAPH_SYNC_HANDOFF_MISSING
```

Both sites now go through one seam, `graph_sync.join_registered_bounded`. Fixing this site-by-site is how the second one survived the first pass, so the guard is structural rather than a third patch: `test_every_graph_join_site_is_bounded_or_declared` enumerates every `wait_for_registered` caller under `src/exomem/` and fails unless each is bounded or on a declared opt-out list with a stated reason. A third site cannot appear silently.

Declared opt-outs, all deliberate: `reconcile.py` (×2), whose terminal asserts graph currency and is the opt-in path; and `epistemic_graph.py` (×2), `delete_file.py`, `delete_directory.py`, `recover_from_trash.py` (×2), all gated on `active_mutation_request_id() is None` **and** no active direct guard — structurally unreachable while a request is being served, with no envelope to carry `pending` and a contract that promises a converged result.

Red on base for the newly-found site:

```
E       AssertionError: direct mutation guard parked 30.0s on a registered rebuild
E       assert 30.04071059997659 < 15.0
```

## The bound

`INTERACTIVE_GRAPH_JOIN_TIMEOUT_SECONDS = 2.0`, sitting just above the latency gate's existing `COMMIT_P95_MS = 1_500.0` so a worst-case write stays the same order of magnitude as the write itself.

Deliberately **not** tuned to "usually be enough." A real full-corpus pass costs 20–175 s, so any bound that sometimes succeeds would sometimes stall a user for minutes. Work that finishes inside 2 s was already nearly done. Not configurable — a knob here is a knob for making writes slow again.

What happens when it is hit:

- **`after_operation_guard`** returns `graph_sync: "pending"` with `GRAPH_SYNC_REBUILD_IN_PROGRESS`, the checkpoint, and remediation.
- **`mutation_guard`** has no terminal envelope, so it logs `direct mutation left its graph rebuild running join_timeout_s=…`. The leaf's own bounded index/readiness paths already tell its caller the graph is not current.
- The helper returns `False` **only** on `TimeoutError`. A genuine `GraphRebuildRegistrationError` still raises, so a failure is never laundered into "pending" — pinned by `test_the_bounded_helper_does_not_launder_a_real_failure_into_pending`.

## Class C converges; the other two are named, not silently skipped

| Class | Addressed |
|---|---|
| **C** — `GraphProjectionMoved`, "the recall projection identity moved across the pass" | **Yes.** A moving-projection attempt re-targets the newer baseline instead of exhausting at 2 attempts. |
| **B** — "the availability marker would not publish" | **No, deliberately.** Retrying a Class B re-pays a doomed pass (#566's finding), and `retarget` is never set on that branch. The supersession subtype is #577's territory, now in this base. |
| **Third variant** — `GraphPublicationUnavailable("another graph rebuild owner did not publish a current sidecar")`, raised at off-boundary entry *before* any stabilization attempt | **No — out of scope, and worth its own lane.** |

That third variant deserves attention: it is **already bounded** (`wait_for_current(..., timeout_seconds=30.0)`), so it is not a latency source. But it means rebuild owner A waited 30 s for owner B's sidecar, B never published, and A then failed too. That is a **second independent contributor** to the marker never being republished, untouched by anything here. Generations 787/788/793/794 all failing inside one hour is consistent with it compounding with Class B.

The bounded join is orthogonal to all three, which is the point: it decouples user-visible latency from a rebuild that on this evidence may simply never converge.

## Stabilization bounds

`REBUILD_STABILIZATION_DEADLINE_SECONDS = 120.0` **and** `REBUILD_STABILIZATION_MAX_ATTEMPTS = 8` — both, because neither alone is a bound. An attempt ceiling is not one when a pass costs 20–175 s (8 passes ≈ 23 min); a deadline is not one when a pass is cheap enough to spin. The existing `REBUILD_STABILIZATION_ATTEMPTS = 2` becomes a **floor**, so this never does fewer attempts than today.

Hitting either bound raises `GraphProjectionMoved` with today's type and today's admitted cause, and the single `mark_external_pending` in the `finally` is untouched. Worst-case composition is `REBUILD_PUBLICATION_ATTEMPTS = 4` × 120 s ≈ 8 minutes — entirely on the background daemon thread, never on a caller.

## Measured

Synthetic corpus with a concurrent writer, timing `LeaseManager.invoke`:

| | median | outcome |
|---|---|---|
| base, 300 pages | 3199.4 ms | `failed` / `GRAPH_SYNC_REBUILD_STOPPED` |
| base, 1200 pages | 4503.1 ms | `failed` / `GRAPH_SYNC_STABILIZATION_EXHAUSTED` |
| **fix, 1200 pages** | **2011.6 ms** | `pending` / `GRAPH_SYNC_REBUILD_IN_PROGRESS` |

One sample came in at 311 ms — the bound is a ceiling, not a floor. Note the outcome change matters as much as the number: base *fails* after 4.5 s, the fix *defers* after 2 s and says so.

Red → green: `11 failed, 3 passed` on `origin/main` → `14 passed` on the commit. The three green-on-base are by design and not claimed otherwise — the fast-path-still-completes test, the call-site enumeration guard, and the broken-projection regression.

## Read-after-write is safe

No read path blocks on a rebuild: `find` falls back to 1-hop wikilink expansion, `graph_context` returns `available: false`, and relation-filtered recall raises the pre-existing typed `RETRIEVAL_INDEX_WARMING` with `retry_after_ms: 250` and schedules a background rebuild. No spec, doc, or test asserts graph-current-at-return — the openspec spec states the opposite. `graph_sync: "completed"` is decided by epoch parity, never by `available()`.

## Two things reviewers should look at

1. **`src/exomem/mutation_terminal.py` is outside the brief's allowlist.** Unavoidable: the compact projection whitelists `graph_sync` values, so `"pending"` would have been silently dropped and a bounded write would be byte-indistinguishable from one with a current graph. One frozenset, used at the two existing sites. This is the same class of defect as PR #582's route — an honest signal that never reaches the default response.
2. **`scripts/records_live_acceptance.py:846` restricts `graph_sync` to `{"completed","failed"}` and will reject `"pending"`.** Not run in CI and outside the allowlist, so it was left alone. Needs a one-line follow-up before that script is used against a bounded build.

## Environmental, not attributable

The read-after-write budget gate fails on this Windows box **on `origin/main` too** — fix `2000: 1361.7ms`, `8000: 4686.5ms / p95 4777.0ms`; base `2000: 971.5ms`, `8000: 4425.1ms / p95 4759.7ms`, with base's first attempt worse still at `read_after_write_p95 89120.1ms`. Same four rows, same shape, fix inside base's spread on every one. Another session was running the full suite throughout and the log carries a `hold_ms=72359.99` boundary warning. Linux CI is the authoritative verdict.

Regression comparison used base-vs-fix **set** difference across 62 modules, one file per process (pytest-timeout's `thread` method kills a whole session, so a single-process run loses everything): base 216, fix 215, **zero test-level differences**.

Rebased onto `4a0c790b` (#577) with one conflict in the constants block, both sides purely additive, both kept. Verified the semantic interaction: #577's `GraphPublicationSuperseded` branch never sets `retarget`, so this extension cannot apply to it, and a mixed run ending in supersession exits at the floor exactly as before.

`test_full_rebuild_twice_moving_vault_is_marked_unavailable` was renamed to `…of_a_continuously_moving_vault_…`; its contract assertions are untouched, only `assert acquisitions == 2` became the new bound since 2 is now the floor.

### DIFF
--- a/src/exomem/epistemic_graph.py
+++ b/src/exomem/epistemic_graph.py
@@ -63,6 +63,26 @@ REBUILD_PUBLICATION_ATTEMPTS = REBUILD_STABILIZATION_ATTEMPTS * 2
 # single pass — while a supersession that survives it is not transient, so
 # re-running the rebuild cannot be what fixes it.
 REBUILD_SUPERSESSION_RETRIES = 1
+
+#: #576. `REBUILD_STABILIZATION_ATTEMPTS` is the attempt *floor*, not the
+#: ceiling: two passes cannot converge against a corpus that is still being
+#: written to, and failing is precisely what strands the availability marker so
+#: the next write falls back into another full rebuild -- a loop that feeds
+#: itself. An attempt invalidated by a moving projection may therefore re-target
+#: the newer baseline instead of exhausting.
+#:
+#: The re-target is bounded by BOTH a wall-clock deadline and an attempt
+#: ceiling, because neither alone is a bound here. An attempt ceiling is not one
+#: when a full-corpus pass costs 20-175 s (8 passes would be ~23 min); a
+#: deadline alone is not one when a pass is cheap enough to spin. 120 s keeps
+#: the worst case (deadline plus the one pass already in flight when it
+#: expires) at or below today's two full passes at production scale, so this
+#: never costs more wall time than the code it replaces -- it only spends it
+#: better. Hitting either bound raises `GraphProjectionMoved` exactly as today:
+#: same Class C type, same admitted cause, and the single
+#: `mark_external_pending` in the `finally` below is untouched.
+REBUILD_STABILIZATION_DEADLINE_SECONDS = 120.0
+REBUILD_STABILIZATION_MAX_ATTEMPTS = 8
 _AVAILABILITY_FRESHNESS_KEY = "recall_projection_identity"
 _RECALL_CHECKPOINT_KEY = "recall_projection_checkpoint"
 _RESOLVER_TOPOLOGY_KEY = "recall_resolver_topology"
@@ -746,6 +766,27 @@ def _recall_projection_identity(
     """
     policy_version, access_fingerprint = recall_policy.recall_policy_identity(vault_root)
     return disk_freshness, policy_version, access_fingerprint
+
+
+def _may_restabilize(attempts: int, *, retarget: bool, started: float) -> bool:
+    """Whether `_rebuild_all_locked` may run another stabilization attempt (#576).
+
+    Three rules, in order:
+
+    * below `REBUILD_STABILIZATION_ATTEMPTS` this is unconditionally True, so a
+      bound can never buy fewer attempts than the code it replaced;
+    * above it, only an attempt invalidated by a *moving projection* earns
+      another one -- a Class B publication refusal is not made truer by
+      repetition (#566);
+    * and that extension stops at whichever of the attempt ceiling and the
+      elapsed deadline comes first, so continuous writes cannot turn the
+      re-target into an unbounded restart loop.
+    """
+    if attempts < REBUILD_STABILIZATION_ATTEMPTS:
+        return True
+    if not retarget or attempts >= REBUILD_STABILIZATION_MAX_ATTEMPTS:
+        return False
+    return (time.monotonic() - started) < REBUILD_STABILIZATION_DEADLINE_SECONDS
 
 
 def _incremental_projection_identity(
@@ -1438,6 +1479,11 @@ class EpistemicGraphIndex:
                             note_publication_refusal(self.vault_root)
                             raise
                         clear_publication_refusal(self.vault_root)
+                        log.info(
+                            "graph rebuild published publication_attempts=%s generation=%s",
+                            attempts,
+                            required.generation if required is not None else None,
+                        )
                         return report
                     finally:
                         _release_publication_hold(publication_hold)
@@ -1467,6 +1513,9 @@ class EpistemicGraphIndex:
         # exception type already carries that classification; only the memo is
         # new, so the next cycle does not re-pay the same doomed rebuild.
         note_publication_refusal(self.vault_root)
+        log.info(
+            "graph rebuild publication exhausted publication_attempts=%s", attempts
+        )
         raise graph_sync.GraphRebuildRegistrationError(
             "GRAPH_SYNC_STABILIZATION_EXHAUSTED",
             # "run reconcile" is an internal registry name that matches neither
@@ -1752,8 +1801,19 @@ class EpistemicGraphIndex:
         # cause belonging to the branch it takes.
         moved_cause = "no stabilization attempt completed"
         publication_cause = "no stabilization attempt completed"
+        # #576. Whether the *last* attempt was invalidated by a moving
+        # projection, which is the only condition worth re-targeting: the
+        # newer baseline is sampled fresh at the top of every attempt, so an
+        # attempt that lost a race to a concurrent write can win the next one.
+        # Class B (the marker would not publish) is deliberately excluded --
+        # retrying it just re-pays a doomed pass, which is #566's finding.
+        retarget = False
+        attempts = 0
+        started = time.monotonic()
         try:
-            for _attempt in range(REBUILD_STABILIZATION_ATTEMPTS):
+            while _may_restabilize(attempts, retarget=retarget, started=started):
+                attempts += 1
+                retarget = False
                 before_disk = _disk_vault_freshness(self.vault_root)
                 before = _recall_projection_identity(self.vault_root, disk_freshness=before_disk)
                 resolver = find_module.recall_resolver_snapshot(
@@ -1772,6 +1832,7 @@ class EpistemicGraphIndex:
                     # Class C, first admitted cause.
                     self._mark_unavailable()
                     projection_moved = True
+                    retarget = True
                     moved_cause = "the supplied freshness identity did not name the resolver bytes"
                     find_module.unload_ram_caches()
                     continue
@@ -1824,6 +1885,7 @@ class EpistemicGraphIndex:
                         # The projection moved between writing the availability
                         # marker and confirming it: Class C, second cause.
                         projection_moved = True
+                        retarget = True
                         moved_cause = (
                             "the recall projection moved after the availability "
                             "marker was written"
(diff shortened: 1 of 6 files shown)


Artexis10:
Superseded — every piece of this landed on main by another route, and the approach itself was replaced rather than merged.

The bound this PR tuned is gone entirely: the join now *polls* (`graph_sync.join_registered_if_settled`), so the interactive write never parks and there is no constant to size. That was the unlandable part — the bound had to be under the 750 ms `COMMIT_MEDIAN_MS` gate *and* long enough for a small test-vault rebuild to converge, which is not a constant that exists.

The contributions that were independently correct are on main already:
- `graph_sync.committed_graph_pending` and the fourth terminal outcome `GRAPH_SYNC_REBUILD_IN_PROGRESS` (`graph_sync.py:3357`, consumed at `writer_lease.py:2603`)
- `tests/test_bounded_graph_join.py`, which still enumerates every join site so a third unbounded one cannot appear silently
- `INTERACTIVE_GRAPH_JOIN_TIMEOUT_SECONDS` deleted rather than retuned

The concurrency property this was reaching for now has an actual gate: `graph convergence (concurrent writer)` runs with `--max-blocking 0` (#605), which asserts no write parks at all rather than asserting a bound.

Closing rather than rebasing — the branch is also conflicting, and rebasing it would only re-land code that is already there.

PR #606, opened 2026-08-17
fix(write-path): let a structural suggestion resolve when its material gets a home

Artexis10:
## Why

A structural-promotion suggestion the user has already acted on never stops firing.

`add-structural-promotion-suggestions` shipped with no resolution mechanism on the stated assumption that the advisory "resolves by scope agreement" — the out-of-scope line in `close-write-warning-suppression`. That holds only for restructure by *subtraction*: the shipped control asserts the origin page is quiet again after "the material removed", and task 1.9 calls it "the cleaned original".

Real restructures are additive. Durable units carry anchors, inbound relation targets and history, so an agent acting on the advice creates the destination pages and leaves the origin units in place. Every gate still holds, and the advisory re-fires on every subsequent write, forever, about material that already has a home. A channel that cannot be resolved teaches its reader to stop reading it, which mutes the true positives too.

Replaying the merged detector against the page that motivated the original feature: it still returns `strength: strong`, `off_scope_units: 16`, all four reason codes — and sixteen of its twenty-one recurring terms are declared by destination pages that exist *because the user acted*.

Pre-registered family `f25 restructure_lifecycle` already measures exactly this and is expected-red: "applying a suggestion resolves it: the signal disappears because the state changed, not because anyone dismissed it."

## What changed

- A cluster is **routed** when eligible compiled pages already declare its vocabulary; routed terms are removed and the **existing** mass gate decides. No second threshold, so a half-routed cluster still reports its unrouted half.
- Resolution comes from corpus state alone. No dismissal, acceptance, snooze or cooldown is recorded — deleting the destinations brings the advice back. `f25` explicitly fails a product that clears its own suggestion by recording a dismissal, so this was a constraint, not a preference.
- A destination must cover at least **two** cluster terms, so an unrelated page carrying one incidental tag cannot scavenge a cluster into silence.
- Destinations come from the corpus the mutation already built. `CreationPreflight` now retains the context `_evaluate_structural` was already building and discarding, so all four compiled writers are covered rather than two — the "invisible feature reaching one writer in four" failure recorded in `exomem-write-feedback-never-reaches-the-default-mutation-response`.
- Destination lookup is restricted to `eligible_compiled_paths`, so a page the caller cannot see cannot change what the caller is told.
- Payload unchanged: no new key, reason code, or numeric quantity. An absent corpus emits exactly as before — suppression is never the fallback.

## Verification

- `tests/test_structure_promotion.py` — 22 passed. All 14 pre-existing tests pass **unchanged**, proving the no-corpus path is identical to today.
- **Mechanism binding**: with `_routed_terms` neutered, exactly the 4 positive resolution tests go red and all 4 negative controls stay green.
- 167 passed across detector, mutation-terminal, mutation-timings, observe and creation-writer suites after merging `main`.
- 293 passed across governance-egress, scaffold-leak, MCP schema fidelity and bootstrap-budget gates. Tool-surface fingerprint unchanged.
- `openspec validate --strict` passes; `ruff check` clean.
- `tests/test_corpus_aware.py` has 4 failures — environmental, a sandbox `PermissionError` on `~/.cache/huggingface`, reproducing identically on clean `origin/main`.

### Reported honestly

The real page does **not** go silent. 16 of 21 terms resolve; it still reports the 5 that genuinely have no home (`countryside, income, ptz, rita, savings` — a household-financing thread the split never routed). That is the intended outcome: silence would require suppressing material with no destination, and clearing on a "mostly routed" ratio would have buried it. An earlier draft of the design doc predicted a full clear by reasoning from the truncated 6-term payload instead of the full 21-term seed set; the document now records the measurement instead.

Cost: the destination pass is deferred until a cluster has formed, so ordinary writes pay nothing. On a write that would otherwise speak it measures 9.2 ms at 2k pages and 33.4 ms at 8k, against a 750 ms median ceiling.

`f25` stays red: it also asserts absence across the due-state counters surface, which S1 has not built. The detector-side claim is reported separately rather than folded into a green.


Artexis10:
## CI status

**`required CI gate`: SUCCESS.** The only remaining red checks are the advisory `cross-platform` (Windows/macOS) shards, which are **not required** and fail identically on clean `origin/main` — verified against main's own run `32053284938`, where all four of the same shards fail.

Two extra commits on this branch are unrelated pre-existing breakage on `main`, not part of the change. `main` was red when this branch was cut:

- `fix(openspec)` — a Windows absolute path in a merged design doc failing the package build's public-input validation.
- `chore(plugins)` — scaffold drift in `references/audit-checks.md` failing the required `test_plugin_sync`, regenerated with the documented command.

Both also appear on #607; whichever lands first makes the other a no-op.

One transient failure is worth naming rather than hiding: `product E2E` failed once with `graph did not converge within 120s ... 'graph sidecar unavailable'`, then passed unchanged on re-run. It passes on `main` and on #607. Given #588, #591 and #592 are all open against exactly this graph-rebuild parking behaviour, I read it as ambient flakiness in that area rather than anything this change introduces — but I could not prove that from one re-run, so it is recorded here for whoever reviews.

Also worth flagging separately: a live governed write during this session returned `graph_sync: failed / GRAPH_SYNC_REBUILD_STOPPED`, which is the same subsystem.

PR #612, opened 2026-08-17
fix(ci): unbreak main by resyncing the plugin and removing a committed local path

Artexis10:
`main` is red. Two independent causes, neither related to the other, both landed around 0.54.1.

## 1. The shipped plugin drifted from the scaffold

`plugins/claude-code/skills/exomem/references/audit-checks.md` no longer matched `src/exomem/_scaffold/_Schema/`, which `CLAUDE.md` names as the single source of the skill. The committed copy was missing the current wording for the unfinished-experiments check and the **entire prediction-window check**.

Regenerated with the command the failing test names:

```
exomem package-skills --plugin-root plugins/claude-code
```

That also carries the plugin manifest version from 0.52.3 to 0.54.1, which had been left behind by the same drift.

Failed: `tests (py3.11, shard 2/4)`, `tests (py3.11, shard 3/4)`, `tests (py3.13, shard 2/4)`, `tests (py3.13, shard 3/4)`.

## 2. An absolute local path was committed in an OpenSpec design

`openspec/changes/bound-corpus-context-flight-join/design.md` cited its evidence as coming from a specific machine path. That trips the `absolute_local_path` rule in the public-repository privacy scan — which is precisely what the rule exists to catch. The sentence keeps its meaning without naming a machine-specific location.

This one failed twice under different names: `package build` runs the same scan as `test_scaffold_no_leak.py`, so a single finding accounted for two red jobs.

## Verification

`pytest tests/test_plugin_sync.py tests/test_scaffold_no_leak.py` → 18 passed (both were failing before).

## Why this is separate

Found while getting #611 green. Keeping it out of that PR so the deployment-lock change stays reviewable on its own — this is a CI fix with no bearing on the lock.


Artexis10:
Superseded by #609, which landed the same two fixes — the absolute local path and the plugin resync — while this was in CI. Verified on `origin/main`: the scaffold and the shipped `audit-checks.md` are now byte-identical, so there is nothing left here to land.

PR #623, opened 2026-08-18
fix(graph): drain queued graph repair on dispatch instead of the 300s reconcile

Artexis10:
> **Draft — correction.** I claimed this PR's own `product E2E` job would be the
> proof. It ran and **disproved it**: the E2E still times out at 120 s with
> `'graph': {'available': False, 'reason': 'graph sidecar unavailable'}`. Two
> reasons, both of which I should have checked before claiming it:
>
> 1. `_record` drops server-authored writes via `_is_self_write_event`, so an
>    MCP write never wakes the dispatch thread this change hooks.
> 2. The E2E sets `EXOMEM_DISABLE_FILE_WATCHER=1`, so in that run there is no
>    watcher at all and *nothing* drains the queue by any path.
>
> What is below is still true and still worth landing — out-of-band edits
> currently wait up to 300 s and after this they do not — but it does **not**
> fix the E2E or the MCP-write case, and this PR should not be merged as if it
> did. Holding it in draft until #625 lands the server log and the real deferral
> reason is visible rather than guessed at.

## The queue was built. Nothing was scheduled to drain it.

Phase 2 shipped the durable per-path `graph_upserts` queue and the incremental
drain that consumes it, and both work. What was never wired is a **scheduler**.
Every call site of the graph drain lives inside `_reconcile_once`, and that runs
from `self._stop.wait(self._reconcile_interval_seconds())` — **300 s** by
default, **900 s** in quiet mode.

So between one tick and the next, an out-of-band edit left the graph owing work
with nothing scheduled to do it: `graph_sync.status()` reported
`recovery_required` and readers got `graph sidecar unavailable`.

## The evidence

#620's diagnostic printed:

```
state='recovery_required'   epoch_kind='coherent'
external_pending=False      graph_queue_depth=13
```

`coherent` is in `REPAIRABLE_EPOCH_KINDS`, so the drain would have admitted this
queue on sight. It held unchanged for the full 120 s the E2E waits.

**What that evidence does and does not establish:** it proves the queue is
non-empty and admissible, and that nothing drained it. It does *not* prove the
cause was the reconcile cadence — in that run the watcher was disabled outright.
The cadence gap is real and this fixes it; the E2E's failure has a different
proximate cause still to be identified from the server log.

## What this changes

The watcher's dispatch thread already runs **off the write path**, so this does
not undo Phase 1 taking the graph off it, and it already **coalesces bursts**, so
a batch of saves settles the graph once rather than once per file. The periodic
reconcile is untouched and stays as the cross-process backstop.

- **The drain runs in a `finally`, on every dispatch** — including the three
  paths where `_flush` returns early. The debt is enqueued by the *write*, not by
  this batch, so a batch with nothing left to dispatch can still have left the
  graph owing work. `_flush`'s body moves to `_dispatch_coalesced_batch` so that
  ordering is explicit rather than implied.
- **It takes the graph queue alone**, via a new `index_sync.drain_graph_work`,
  rather than `drain_deferred_work`. The reconcile wants all three queues because
  it runs every 300 s; this seam fires within a debounce window, and replaying
  embeddings that often is a different cost decision.

## Verification

- Graph, index-sync, reconcile, watcher and deferred-drain suites run against
  this branch **and** against an `origin/main` baseline worktree produce
  **byte-identical failure sets** (`comm` diff empty both directions; 37
  pre-existing Windows DACL failures either side), with 4 new tests on top.
- New tests: the drain fires on an empty batch with the policy limit; it still
  fires when the batch dispatch raises; a failing drain never kills the dispatch
  thread; `drain_graph_work` leaves the semantic/full queues alone.
- `ruff check` clean.

### DIFF
--- a/src/exomem/file_watcher.py
+++ b/src/exomem/file_watcher.py
@@ -377,8 +377,48 @@ class FileWatcher:
         return media, ups, del_rels, pending_epoch, access_policy
 
     def _flush(self) -> None:
-        """Dispatch the coalesced batch: publish freshness/inbound for every
-        changed path (vault-wide), and re-embed only the Knowledge Base subset."""
+        """Dispatch the coalesced batch, then settle whatever graph debt it left.
+
+        The drain runs in a `finally` and on every dispatch, including the three
+        paths that return early: graph debt is enqueued by the write itself, so
+        a batch with nothing left to dispatch here can still have left the graph
+        owing work.
+        """
+        try:
+            self._dispatch_coalesced_batch()
+        finally:
+            self._drain_graph_debt()
+
+    def _drain_graph_debt(self) -> None:
+        """Repair the graph within a debounce window of the write that dirtied it.
+
+        Deferred graph work used to drain only from `_reconcile_once`, which runs
+        on `reconcile_interval_seconds` -- 300s by default, 900s in quiet mode.
+        Until that tick fired, `graph_sync.status()` reported `recovery_required`
+        and readers got `graph sidecar unavailable`, so an ordinary write left the
+        graph unreadable for up to five minutes with the repair already queued,
+        already admissible, and nothing scheduled to run it. The product E2E
+        proved it: `state='recovery_required' epoch_kind='coherent'
+        external_pending=False graph_queue_depth=13`, held for the whole 120s the
+        loop waits -- a wait that could not succeed against a 300s cadence.
+
+        The dispatch thread is the right seam. It already runs off the write path
+        (so this does not undo taking the graph off it) and it already coalesces
+        bursts, so a batch of saves settles the graph once rather than per file.
+        The periodic reconcile stays as the cross-process backstop for debt this
+        process never observed.
+        """
+        try:
+            index_sync.drain_graph_work(
+                self._vault_root,
+                limit=self._watcher_policy().max_reconcile_embed_files,
+            )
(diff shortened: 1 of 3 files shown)


Artexis10:
Superseded — closing in favour of the graph drain daemon (PR to follow from
`feat/graph-drain-worker`).

This PR hooked the drain to the **file watcher's dispatch thread**. Two things
found since make that the wrong seam, both established from the server's own log
rather than by reading:

1. `_record` drops server-authored writes via `_is_self_write_event`, so an MCP
   write never wakes that thread at all — the case that matters most was the one
   case this could not reach.
2. The E2E sets `EXOMEM_DISABLE_FILE_WATCHER=1`, and the watcher also no-ops
   without `watchdog`. So convergence was being attached to an **optional**
   component; where it is absent nothing drained the queue by any path.

The replacement owns the schedule directly: a daemon started by the server
runtime, woken by a signal on every graph enqueue (`add_graph` /
`mark_graph_full_rebuild`, both reached by `enqueue_graph_checkpoint`, which
every canonical batch calls), debounced so it drains the quiet moment *after* a
write burst instead of into the middle of one, and backed off so a queue that
cannot drain costs one attempt every two minutes rather than one per second.

The two pieces worth keeping from here are carried over intact: the
`index_sync.drain_graph_work` seam that takes the graph queue alone, and the
reasoning about why the periodic reconcile stays as the cross-process backstop.

PR #631, opened 2026-08-18
fix(tests): declare what the Linux-only benchmark harnesses require

Artexis10:
## These harnesses are Linux-only on purpose

`benchmarks/memorybench` and `benchmarks/epistemic` are not product code and were never meant to be portable. Their contracts are written *in terms of* Linux facilities, because reproducible benchmark evidence is the point of them:

- `memorybench.export._secure_read` proves a private file is unreadable by anyone else, using POSIX mode bits and `st_uid`.
- `memorybench` pins Bun to an exact version so a run is byte-reproducible.
- `epistemic.broker` bounds provider work with `setitimer`/`SIGALRM` and isolates it with `bwrap` — Linux user namespaces, which exist nowhere else.

`ci.yml` provisions all three, and every job in it is `ubuntu-latest`. `cross-platform.yml` cannot, so it collected **135 failures and 2 errors** that say nothing about the code. On Windows `Path.chmod(0o600)` only toggles the read-only bit and the file still reports `0o666`, so the very first `_secure_read` of a test's own plan raised before the test reached its subject.

A missing prerequisite is a skip. This declares the prerequisites.

## The more interesting half: 35 false passes removed

A test asserting that *its injected fault* yields `BLOCKED` was green on Windows because the **permission** failure yielded `BLOCKED` first. Every one of these was passing without testing anything:

- all 7 `test_every_invalid_run_plan_is_a_quiet_blocked_result` cases
- all 7 `test_preflight_refuses_identity_drift_as_blocked_without_provider_work` cases
- all 5 `test_native_dataset_and_fresh_runtime_state_block_before_any_stage` cases
- all 3 `test_preflight_resolves_and_verifies_bun_and_uv_outside_os_defpath` cases
- all 13 `test_verify_refuses_*` — a `pytest.raises` satisfied by `"Bun runtime is unavailable"` rather than by the drift each one injects

None could tell its own fault from the platform's. A skip that names the absent capability is worth more than a pass that proves nothing.

## How the gates are placed

Gated on the **capability**, never on `sys.platform`, so a Linux box without Bun also reports the truth and the reason names what is actually missing. Each gate sits on the narrowest seam that needs it rather than on a whole module, so the pure-logic tests — digest vectors, plan validation, canonical selection — keep running everywhere. That is the cross-platform signal worth having.

The broker takes the `conftest` route instead: its 19 failing entry points share no helper to gate, and it already refuses up front with a self-describing contract error, so the hook #626 added grows one branch. It matches through `__context__` as well, because `pytest.raises(..., match=...)` re-raises as an `AssertionError` holding the refusal — 8 of the 19 failed exactly that way.

## Two things fixed rather than skipped

Both are defects on any platform:

- **`test_driver_ipc_rejects_...oversized_messages`** had a 64KiB payload as an unnamed parametrize case, so the payload *became the test id* and pytest exported the node id as `PYTEST_CURRENT_TEST` — past the 32767-char cap Windows puts on an environment variable, erroring in setup and teardown without ever running the body. All seven cases are named now, which reads better everywhere.
- **`test_cli_has_strict_plan_only_surface_and_no_fixture_fault_switch`** now builds its own path instead of a 0600 plan. Argument parsing rejects its arguments before the plan is ever opened, so it stays a real test everywhere — the one genuine pass the gates would otherwise have cost.

## Measured on Windows

`-p no:randomly`, whole modules:

| module | before | after |
|---|---|---|
| `test_memorybench_export` | 89 failed, 42 passed | **0 failed**, 20 passed, 112 skipped |
| `test_epistemic_broker` | 30 failed, 13 passed, 2 errors | **0 failed**, 13 passed, 31 skipped |
| `test_memorybench_setup` | 16 failed, 15 passed | **0 failed**, 2 passed, 29 skipped |

Every pass that became a skip is one of the 35 false passes enumerated above; all 35 genuine passes are still passes.

Linux behaviour is unchanged by construction — every gate is a no-op where the capability exists, and every `conftest` branch is guarded on the capability being absent, so a `CustodyUnsupported` or a timer refusal on Linux stays a failure and is meant to. **CI's ubuntu lanes are what confirm that**, not the local Windows run.

### DIFF
--- a/tests/benchmark_capabilities.py
+++ b/tests/benchmark_capabilities.py
@@ -35,6 +35,7 @@ import shutil
 import signal
 import subprocess
 from functools import lru_cache
+from pathlib import Path
 
 import pytest
 
@@ -92,10 +93,30 @@ def has_posix_interval_timers() -> bool:
     return all(hasattr(signal, name) for name in _TIMER_PRIMITIVES)
 
 
+#: What `epistemic.broker` pins its sandbox to. Not merely "Linux": these are
+#: Debian/Ubuntu x86-64 paths, hardcoded, and the broker refuses rather than
+#: substitute -- it verifies `bwrap` *is* the trusted binary, then binds this
+#: exact runtime into the namespace. Duplicated here rather than imported so a
+#: skip decision never depends on importing the harness it is gating.
+_SANDBOX_RUNTIME = (
+    "/usr/bin/bwrap",
+    "/usr/bin/python3.12",
+    "/usr/lib/python3.12",
+    "/lib/x86_64-linux-gnu",
+    "/lib64/ld-linux-x86-64.so.2",
+)
+
+
 @lru_cache(maxsize=1)
-def has_bwrap() -> bool:
-    """True when a bubblewrap sandbox can be spawned at all."""
-    return shutil.which("bwrap") is not None
+def has_bwrap_sandbox() -> bool:
+    """True when the exact sandbox the broker binds is present.
+
+    `shutil.which("bwrap")` alone is not enough: the broker resolves it against
+    `/usr/bin/bwrap` specifically and binds a pinned interpreter, stdlib, libc
+    and loader beside it. Any one of them missing is a refusal, so all of them
+    are the capability.
+    """
+    return all(Path(path).exists() for path in _SANDBOX_RUNTIME)
 
 
 def require_posix_file_modes() -> None:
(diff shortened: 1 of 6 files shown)


Artexis10:
### Evidence the gates are capability-conditional, not blanket skips

The worry with a skip-based fix is that it quietly removes tests that were passing. Forcing every predicate in `benchmark_capabilities` to return `True` — the state Linux CI is actually in — and re-running all three modules on this Windows box:

```
107 failed, 70 passed, 30 skipped
```

**70 passed** is exactly the baseline count on `origin/main` here (42 export + 13 broker + 15 setup = 70). Not one passing test is removed by a gate; every gate is a no-op wherever its capability exists.

The 107 failures in that forced run are expected and are *not* the claim — forcing a predicate does not conjure Bun, `bwrap`, or POSIX mode bits into existence, so those tests still fail on Windows for the real reason. The measurement that matters is the pair: gates off → the same 70 pass; gates on → 0 fail and 35 of those 70 become skips, all 35 individually enumerated as false passes in the description.

Linux is confirmed by this PR's own ubuntu lanes, not by anything local.

PR #659, opened 2026-08-19
fix(windows): give the membench hook harness an environment a child can run in

Artexis10:
Third-largest cause on the Windows lane: 14 failures across `test_membench_trackc` and `test_membench_trackd`, every one of them reported as

```
RuntimeError: install-hook --client claude failed rc=1: Traceback (most recent call last):
  File "src/exomem/__main__.py", line 33, in <module>
    import asyncio
  ...
  File ".../asyncio/windows_events.py", line 8, in <module>
    import _overlapped
OSError: [WinError 10106] The requested service provider could not be loaded or initialized
```

## What was wrong

The track C harness built its child environment as a fresh dict holding a hardcoded `PATH = "/usr/bin:/bin"` and five exomem variables. On Windows that is not isolation, it is an environment no process can start in: Winsock resolves its service providers underneath `SystemRoot`, so a child without it dies importing `asyncio` before any exomem code runs.

Four more platform assumptions sat behind that one, each invisible until the one in front was fixed:

**1. `install_hook` computed its defaults with `Path.home()` at module scope.** Windows resolves `~` from `USERPROFILE`, not `HOME`, so a child given only `HOME` has no home at all and `Path.home()` raises — at *import*, so `exomem install-hook` died before parsing its own arguments, including invocations that name every path explicitly and never need a default. This is the one product-side fix here: a Windows service account has no `USERPROFILE` either, so this was reachable outside the harness. The defaults now resolve on demand through a module `__getattr__` (every existing name keeps working), and the single comparison that consumed them became `_is_conventional_hook_dir`, which answers *no* where there is no home rather than raising — correctly, since `~` has nothing to expand to and the absolute path is the only form that can work.

**2. `subprocess.run(["bash", ...])` does not consult PATH the way it appears to.** `CreateProcess` searches the system directory *before* any PATH entry, so wherever WSL is installed `bash` is the launcher in System32 — which cannot open the `C:` path the wrapper is invoked with. The hook exited 127 and the suite scored a missing interpreter as a hook that chose not to fire. `bash_executable()` now names a Windows-native bash.

**3. The wired command spells a custom hook directory as a POSIX path**, by design (`_command_for`), so comparing it against the native spelling made every wired entry look as though it pointed outside the isolated home — on a separator alone. And Codex entries carry `commandWindows` precisely because `python3` is not a name Windows has, so the checkpoint driver now takes it there, exactly as a Codex client does.

**4. The injection ladder degraded its CLI rung by dropping the shim directory from PATH.** That was the whole degradation on POSIX, where the base PATH is `/usr/bin:/bin` and never held an `exomem`. On Windows the base PATH must carry an interpreter directory — the wrapper exits 0 with no output when neither `python3` nor `python` resolves, which the suite reads as a hook staying quiet — and for a checkout installed into a venv that same directory holds `exomem.exe`. So the shim came off PATH and the hook resolved the *real* CLI: the ladder never degraded and the case asserted the opposite of what it ran. Entries are now removed by capability (`shutil.which("exomem", path=entry)`) rather than by name, which keeps the interpreter and takes the CLI on either platform.

## Verification

Windows, this box:

```
tests/test_membench_trackc.py    before: 14 failed, 0 passed     after: 14 passed
tests/test_membench_trackd.py    before:  5 failed, 6 passed     after:  5 failed, 6 passed
```

Track D's 5 are pre-existing and identical on `main` at this commit — unrelated causes, left visible.

`tests/test_install_hook.py`: 91 passed. The 4 failures are the same 4 that fail on `main` on this machine (a local `bash` resolution quirk), unchanged by this branch.

POSIX behaviour is unchanged by construction: `platform_env()` returns the same `{"PATH": "/usr/bin:/bin"}`, `home_env()` the same `{"HOME": ...}`, `bash_executable()` the same bare `"bash"`, `as_posix()` the same string as `str()`, and `commandWindows` is consulted only under `os.name == "nt"`.

### DIFF
--- a/benchmarks/membench/trackc/checkpoint_driver.py
+++ b/benchmarks/membench/trackc/checkpoint_driver.py
@@ -47,15 +47,19 @@ from __future__ import annotations
 
 import hashlib
 import json
+import os
+import shutil
 import subprocess
 from dataclasses import dataclass, field
 from pathlib import Path
 
 from membench.trackc.hook_home import (
+    bash_executable,
     HookHome,
     create_hook_home,
     ensure_isolated,
     make_workdir,
+    platform_env,
 )
 
 SCHEMA_VERSION = 1  # exomem_continuation_checkpoint.py line 26
@@ -67,6 +71,25 @@ _GIT_IDENTITY = {
     "GIT_COMMITTER_NAME": "Membench",
     "GIT_COMMITTER_EMAIL": "membench@example.invalid",
 }
+
+
+
+def _git_env() -> dict[str, str]:
+    """Isolated git env that can still find and run git on this platform.
+
+    The identity is pinned so a commit here never depends on the developer's
+    global config. `platform_env` supplies the rest; on Windows a child given
+    only a POSIX `PATH` cannot resolve `git` at all, and cannot start a process
+    that opens a socket either.
+    """
+    environment = {**platform_env(), **_GIT_IDENTITY}
+    git = shutil.which("git")
+    if git is not None:
+        environment["PATH"] = os.pathsep.join(
+            [str(Path(git).parent), environment.get("PATH", "")]
+        ).strip(os.pathsep)
+    return environment
+
 
 WORKSPACE_NAME = "membench-workspace"
 BRANCH_MARKER = "membench-marker-branch"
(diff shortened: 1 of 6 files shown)


Artexis10:
Superseded. #642 repaired the membench track C harness independently — `tests/test_membench_trackc.py` is 5 passed / 9 skipped / 0 failed on `main` at 4df95d55, so the 14 Windows failures this PR targeted are gone.

The one part not covered there was the `install_hook.py` change that stopped four module-scope `Path.home()` calls from making the module unimportable. That is now #666, which takes it further: the same defect was also in `writer_lease.LeaseConfig.state_dir` and `install_skill.py`, and `writer_lease` is in the `exomem.commands` import chain — so the whole command surface was unimportable, not just the hook installer. Tracked as #665.

PR #660, opened 2026-08-19
fix(hosted): declare the runtime temp's POSIX isolation instead of failing on it

### LINKED ISSUE #657: Windows: hosted runtime temp enforces privacy with POSIX uid/gid/mode only, so every hosted-gateway test fails with an opaque HOSTED_TRANSFER_UNAVAILABLE
On the Windows lane, 24 tests across `test_hosted_private_routes`, `test_hosted_transfer_v2`, `test_hosted_adoption_staging` and `test_governance_tokens` fail identically at hosted server construction:

```
exomem.hosted_gateway.HostedGatewayError: HOSTED_TRANSFER_UNAVAILABLE: hosted runtime temp is unavailable
src\exomem\server_hosted.py:874
```

That is the second-largest cluster on the Windows lane — 24 of 241 failures on `main` (run 32223246719).

## Cause

`hosted_runtime_temp._private_directory` enforces privacy with POSIX ownership and mode alone:

```python
path.mkdir(mode=0o700, parents=False, exist_ok=True)
value = path.lstat()
if ... or value.st_uid != expected_uid or value.st_gid != expected_gid:
    raise OSError("runtime temp directory has unsafe ownership")
path.chmod(0o700, follow_symlinks=False)
if stat.S_IMODE(path.lstat().st_mode) != 0o700:
    raise OSError("runtime temp directory is not private")
```

On Windows `mkdir(mode=0o700)` does not produce mode `0o700` for a directory, so the final check fails, the `OSError` becomes `HostedRuntimeTempUnavailable`, and the gateway reports `HOSTED_TRANSFER_UNAVAILABLE`. The module has no `os.name == "nt"` branch anywhere; `_open_root` and `_measure_directory` carry the same uid/gid/mode assumptions, and `hosted_restore.py` uses `os.getuid()` and `os.fchown` directly.

The failure message names none of this. `HOSTED_TRANSFER_UNAVAILABLE: hosted runtime temp is unavailable` is what a reader gets for a platform mismatch, a real ownership problem, and a quota breach alike — the `from exc` cause is dropped at the gateway boundary.

## The question this needs a decision on

The hosted runtime temp is only reachable through a `HostedCellConfig` — `server_hosted.py` and `server_runtime._cleanup_hosted_transfer_temp`, both of which require `runtime_uid` / `runtime_gid` tenant bindings. A personal Windows deployment never constructs one. So this looks like a Linux-container capability whose tests are being run on a platform it does not target, rather than a Windows defect users can hit.

If that is right, the fix is to say so rather than to build a second privacy model:

- Have `ensure_hosted_runtime_temp` / `prepare_hosted_runtime_temp` refuse on Windows with a message that names the reason (POSIX ownership semantics are the isolation mechanism), instead of an opaque "unavailable".
- Gate the affected suites POSIX-only with that reason, the way `test_install_hook.py` already gates its ownership and mode trust tests.

The alternative — implementing a Windows DACL privacy model for `hosted_runtime_temp` mirroring `mutation_lock`'s — adds security surface to a component that will not run in production on Windows, so it should not be done by default.

Note this is a *different* root cause from the 29 `WindowsRuntimeDaclError` failures in the same lane: those are in `mutation_lock`, which does run on Windows in production, and are being fixed separately.

Artexis10:
Closes #657.

Second-largest cause on the Windows lane: 24 of 241 failures on `main`, every one of them at hosted server construction with

```
exomem.hosted_gateway.HostedGatewayError: HOSTED_TRANSFER_UNAVAILABLE: hosted runtime temp is unavailable
```

## What the module actually requires

Tenant isolation in `hosted_runtime_temp` **is** POSIX ownership and mode — every entry is checked against an expected uid and gid and required to be `0700`. A hosted cell is a Linux container by construction: the module is only reachable through a `HostedCellConfig`, which is what carries the `runtime_uid` and `runtime_gid` those checks compare against. A personal deployment never constructs one.

Windows has no equivalent in this module, and `mkdir(mode=0o700)` does not produce mode `0o700` for a directory there, so `_private_directory`'s final check failed and the `OSError` became `HostedRuntimeTempUnavailable`.

Building a second, Windows-specific privacy model for a component that will not run in production on Windows would add security surface that nothing exercises. The honest answer is to say what the capability requires.

## The change

- `hosted_runtime_temp` refuses up front, by name, with `POSIX_PRIVACY_REFUSAL`. That is worth having on its own: `HOSTED_TRANSFER_UNAVAILABLE: hosted runtime temp is unavailable` was the message for a platform mismatch, a real ownership breach *and* a blown quota alike, and the gateway drops the `from exc` cause at its boundary — so a reader had no way to tell them apart.
- `tests/conftest.py` turns that named refusal into a skip on Windows, as the seventh entry in the capability-declaration hook that already handles proc-fd custody, the hosted cell's POSIX APIs, interval timers, the bubblewrap sandbox, trusted system Git, and directory descriptors.

Matched on the **refusal text**, not on the exception type: `HostedRuntimeTempUnavailable` is also how a genuine ownership breach, an escaped symlink and an exceeded quota are reported, and those stay failures wherever they occur. Gated on `os.name == "nt"`, so the same refusal on Linux or macOS — where the semantics exist and their absence would be a real defect — stays a failure and is meant to.

## Verification

`tests/test_hosted_private_routes.py` on Windows:

```
main    27 failed,  2 passed, 1 skipped
branch   0 failed,  5 passed, 25 skipped
```

Three of those 27 were passing tests that had been failing on the shared construction path, so coverage went *up*, not down.

`test_hosted_transfer_v2`, `test_hosted_adoption_staging` and `test_governance_tokens` are unchanged on this box (31 failures on both sides, set difference empty in each direction) because their remaining failures are all `WindowsRuntimeDaclError` — the separate cluster #658 fixes. The two are complementary; neither masks the other.

POSIX is untouched by construction: `posix_privacy_available()` is True there, so the new guard is a no-op, and the conftest branch never runs.

### DIFF
--- a/src/exomem/hosted_runtime_temp.py
+++ b/src/exomem/hosted_runtime_temp.py
@@ -24,6 +24,33 @@ class HostedRuntimeTempUnavailable(RuntimeError):
     code = "HOSTED_RUNTIME_TEMP_UNAVAILABLE"
 
 
+#: Named, because "unavailable" was covering three unrelated conditions.
+#: Tenant isolation here *is* POSIX ownership and mode: every entry is checked
+#: against an expected uid and gid and required to be `0700`, and a hosted cell
+#: is a Linux container by construction -- it is only ever reached through a
+#: `HostedCellConfig`, which carries the `runtime_uid` and `runtime_gid` those
+#: checks compare against. Windows has no equivalent in this module, and
+#: `mkdir(mode=0o700)` does not produce mode `0o700` for a directory there, so
+#: the mode check failed and the gateway reported `HOSTED_TRANSFER_UNAVAILABLE`
+#: -- the same message it gives for a real ownership breach or a blown quota,
+#: with the cause dropped at the boundary. Refusing up front says which of the
+#: three it is.
+POSIX_PRIVACY_REFUSAL = (
+    "hosted runtime temp requires POSIX ownership and mode semantics; "
+    "a hosted cell runs as a Linux container"
+)
+
+
+def posix_privacy_available() -> bool:
+    """Whether this platform has the ownership semantics the isolation needs."""
+    return os.name != "nt"
+
+
+def _require_posix_privacy() -> None:
+    if not posix_privacy_available():
+        raise HostedRuntimeTempUnavailable(POSIX_PRIVACY_REFUSAL)
+
+
 def _private_directory(path: Path, *, expected_uid: int, expected_gid: int) -> None:
     try:
         path.mkdir(mode=0o700, parents=False, exist_ok=True)
@@ -100,6 +127,7 @@ def prepare_hosted_runtime_temp(
 ) -> Path:
     """Create and no-follow clear the disposable runtime root during locked startup."""
 
+    _require_posix_privacy()
     root = ensure_hosted_runtime_temp(
         state_root,
         expected_uid=expected_uid,
@@ -121,6 +149,7 @@ def ensure_hosted_runtime_temp(
 ) -> Path:
     """Validate or create the private runtime root without deleting any entry."""
 
+    _require_posix_privacy()
     state = Path(state_root)
     if not state.is_absolute():
         raise HostedRuntimeTempUnavailable
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -141,6 +141,33 @@ def _needs_an_absent_posix_api(error: BaseException | None) -> bool:
     return False
 
 
+#: The hosted cell's runtime temp refusing a platform it cannot isolate on.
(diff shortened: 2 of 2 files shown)


Artexis10:
Superseded by #642. Its conftest branch is strictly broader than this one: it matches `HostedRuntimeTempUnavailable` by type alongside the security and DACL refusals, adds the hosted-operator mode refusals, and gates on `has_posix_file_modes()` rather than `os.name`. `tests/test_hosted_private_routes.py` is 2 passed / 28 skipped on `main` at 42c94959 — the 27 failures this PR targeted are declared, not failing. Issue #657 is answered there rather than here.

PR #661, opened 2026-08-19
fix(portability): make the quiesced export work on a non-POSIX filesystem

Artexis10:
`export_quiesced_vault` could not complete on Windows at all — 16 of 241 failures on the cross-platform lane, in **three** separate defects that each hid the next. Fixing one only revealed the following one.

## 1. Descriptors opened in text mode

`_open_regular_source` built its flags as `os.O_RDONLY` plus `O_NOFOLLOW` where available. Windows opens a descriptor in **text mode** by default, so every CRLF read back as a bare LF:

```
on disk    20  b'line one\r\nline two\r\n'
os.read    18  b'line one\nline two\n'
+O_BINARY  20  b'line one\r\nline two\r\n'
```

The size check is the only reason this surfaced as a failure rather than as silence. **A manifest is a promise about bytes**, and had the sizes ever agreed, the export would have published a SHA-256 of text the file does not contain.

The same flag was missing at five other descriptor reads — `governance/lifecycle`, `hosted_operator`, `hosted_restore`, `hosted_runtime`, `hosted_security` — all fixed, since it is one defect class. `getattr(os, "O_BINARY", 0)` is `0` on POSIX, so none of it changes anything there.

## 2. A cross-API stat comparison on a field Windows caches

The guard against a file changing between the path stat and the handle stat compared the full signature, including `st_ctime_ns`. On Windows that is the **creation** time, and it is served from the directory entry, which the filesystem updates lazily. A file written moments ago:

```
lstat  (…, 57, 1787130382508331800, 1787130382507331400)
fstat  (…, 57, 1787130382508331800, 1787130382508331800)
```

A millisecond apart, with nothing having touched it.

`_comparable_across_stat_apis` now drops that field on Windows only, where it can detect nothing the others miss: a new inode moves `st_dev`/`st_ino`, a content change moves `st_size`/`st_mtime_ns`, and creation time cannot change for a live inode. On POSIX it is the inode-change time — it catches a chmod, a chown, a link-count change — and stays.

The narrowing applies **only** across the two APIs. The before-and-after-read comparison is untouched: both sides come from `fstat`, and every field there means something.

## 3. Durability flushed through a read-only handle

`_fsync_regular_file` opened `"rb"`. `os.fsync` on Windows is `_commit`, which needs write access and returns `EBADF` without it, so a completed export failed `ARTIFACT_PUBLICATION_FAILED: completed artifact could not be made durable` — on a file it had written seconds earlier. POSIX accepts a read-only descriptor, which is why this was correct until it ran somewhere else.

## Verification

`tests/test_hosted_portability.py` on Windows:

```
main    16 failed, 61 passed
branch   0 failed, 80 passed
```

Three of those 80 are new:

- a CRLF source is digested **exactly as it sits on disk** — asserting the recorded `sha256` and `size` against the raw bytes, so the wrong-digest failure mode is pinned rather than just the refusal it happened to cause;
- creation time is the only field the cross-API comparison drops, and only on Windows;
- a completed artifact is reachable and non-empty.

`test_hosted_operator`, `test_hosted_security` and `test_governance_receipt_runtime_bootstrap_windows`: set difference empty in both directions against `main` (the 10-vs-11 count moves run to run on a known-flaky concurrent bootstrap test).

### DIFF
--- a/src/exomem/governance/lifecycle.py
+++ b/src/exomem/governance/lifecycle.py
@@ -628,7 +628,10 @@ def _read_json(vault_root: Path, path: Path) -> dict[str, Any] | None:
             "LIFECYCLE_PATH_UNSAFE", "lifecycle marker is not a regular file"
         )
     try:
-        fd = os.open(path, os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0))
+        fd = os.open(
+            path,
+            os.O_RDONLY | getattr(os, "O_BINARY", 0) | getattr(os, "O_NOFOLLOW", 0),
+        )
     except OSError as exc:
         raise LifecycleError(
             "LIFECYCLE_PATH_UNSAFE", "lifecycle marker could not be opened safely"
--- a/src/exomem/hosted_operator.py
+++ b/src/exomem/hosted_operator.py
@@ -593,7 +593,12 @@ def read_offline_request(
     ):
         _fail()
     _reject_symlink_components(supplied)
-    flags = os.O_RDONLY | getattr(os, "O_CLOEXEC", 0) | getattr(os, "O_NOFOLLOW", 0)
+    flags = (
+        os.O_RDONLY
+        | getattr(os, "O_BINARY", 0)
+        | getattr(os, "O_CLOEXEC", 0)
+        | getattr(os, "O_NOFOLLOW", 0)
+    )
     try:
         descriptor = os.open(supplied, flags)
     except OSError:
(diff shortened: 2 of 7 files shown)


Artexis10:
Superseded by #642, which landed all three repairs independently: `O_BINARY` on the source descriptor (`hosted_portability.py:563`), the cross-stat-API narrowing that drops `st_ctime_ns` (`:524`), and `r+b` for the durability flush (`:897`). `tests/test_hosted_portability.py` is 77/77 on `main` at 42c94959.

PR #662, opened 2026-08-19
test(writes): stop telling the product it has a kernel feature the platform lacks

Artexis10:
16 of the 22 `test_transactional_writes` failures on the Windows lane, and none of them were about the product.

## What the tests were doing

They replace `os.open`, `os.stat` or `os.scandir` with an observer, then call `_replace_capability_member` to re-register the observer in `os.supports_dir_fd` / `os.supports_fd` so the product keeps taking its descriptor-relative branch.

That set is not a preference. It is the platform reporting a kernel feature. On Windows it is **empty**, so rewriting it told the product it had `openat` and sent it straight into

```
os.open(name, flags, dir_fd=parent_descriptor)
NotImplementedError: dir_fd unavailable on this platform
```

`vault.py` handles Windows perfectly well here — `_open_batch_residue_directory` checks `os.open in os.supports_dir_fd` and falls back to a path open. The tests were failing on a branch the product only ever takes because they forced it to.

## The change

`_replace_capability_member` skips when it is asked to *add* a member for a capability this platform does not have.

**Removing** a member stays available everywhere — a capability that is already absent is exactly the state those tests want, which is why `test_directory_census_classifies_residue_through_path_fallbacks` and its siblings still run on Windows and still cover the fallback there. That asymmetry is the whole of the fix.

## Verification

```
main    22 failed, 42 passed,  2 skipped
branch   6 failed, 47 passed, 13 skipped
```

Eleven became honest skips; **five now pass** — they had been failing on the forced branch and are fine on the real one. The remaining six fail identically on `main` for unrelated causes (`DID NOT RAISE PermissionError`, timestamp assertions) and stay visible; set difference against `main` is empty in both directions.

### DIFF
--- a/tests/test_transactional_writes.py
+++ b/tests/test_transactional_writes.py
@@ -52,7 +52,24 @@ def _replace_capability_member(
     original: object,
     replacement: object | None,
 ) -> None:
+    """Re-register a capability under a patched function, where it exists.
+
+    These tests replace `os.open`, `os.stat` or `os.scandir` and then need the
+    product to keep taking its descriptor-relative branch, which it selects by
+    membership in `os.supports_dir_fd` / `os.supports_fd`. Rewriting that set
+    cannot conjure the kernel feature. On Windows it is empty, and telling the
+    product otherwise sent it into `os.open(..., dir_fd=...)` for a
+    `NotImplementedError` -- against a platform the product itself handles,
+    through exactly the path fallback these tests are not about. Seven failures
+    on the cross-platform lane were that, reported as defects in code that has a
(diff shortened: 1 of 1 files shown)


Artexis10:
Superseded by #642, which landed the same guard in `_replace_capability_member` — skipping when a replacement is installed for a capability the platform does not have, rather than fabricating `os.supports_dir_fd` and driving the production branch into `NotImplementedError`. `tests/test_transactional_writes.py:96-104` on `main` at 42c94959 carries it. The 8 failures still in that file are a separate cluster and are being handled on their own.

PR #664, opened 2026-08-19
fix(census): pin an enumerated entry to an identity that exists

Artexis10:
`adopt(mode="scan-only")` reported an **empty vault on Windows** — zero markdown files, every governance count at zero, `saved_contracts: partial`. Nothing raised and nothing warned; the report was coherent and wrong. Two defects, the second invisible until the first was fixed.

## 1. The identity was pinned to a record that carries no identity

The census pinned each enumerated entry with `os.DirEntry.stat()`. That answers from the directory-enumeration record, and on Windows the record carries no volume serial and no file index — `st_dev` and `st_ino` both come back `0`:

```
DirEntry.stat()  st_dev 0           st_ino 0
Path.lstat()     st_dev 2633241797  st_ino 48132221025888579
```

Every re-check in the module reads them through `lstat` or `fstat`, which open the object and get the real pair, so a pinned directory could **never** match itself. The census counted every subdirectory as having changed underneath it and descended into none of them.

`semantic_census.py:591` was the only DirEntry-derived identity in the product — it fed both the ancestor chain and the per-file identity, so directories and files were both affected. It now takes one real look where the identity is pinned. POSIX fills both fields during enumeration, so the record is returned unchanged and no extra call is made there.

## 2. Then the read compared two stat APIs that disagree

Once the walk descended, every page came back `unsafe`. The post-read check compares the `fstat` taken on its own open handle against an `lstat` on the path, via `_same_file_state` — which includes `st_ctime_ns`. On Windows `st_ctime` is the *creation* time, served from the directory entry the filesystem updates lazily, so those two calls differ by a millisecond for a file nothing touched.

This is the same defect class as #661 (`hosted_portability`), at a second site. The narrowing stays narrow:

- **handle-to-handle** (`_same_file_state(opened, after)`) keeps every field, on both platforms — both sides come from the same API.
- **handle-to-path** drops only `st_ctime`, only on Windows. A new inode still moves `st_dev`/`st_ino`; a content change still moves `st_size`/`st_mtime_ns`.
- **POSIX keeps `st_ctime` everywhere**, where it is the inode-change time and catches a chmod or link-count change the other fields miss.

## 3. One test-side defect went with them

`test_adopt_semantic_census_governance_uses_exact_scanned_page_state` wrote its page with `write_text`, so on Windows it laid down CRLF and then demanded the census hand back LF. A test about exact bytes has to write exact bytes.

## Verification

`tests/test_adopt.py` on Windows:

| | failed | passed |
|---|---|---|
| `origin/main` | 10 | 36 |
| this branch | 0 | 49 |

Three added tests pin the invariants: that an enumerated entry's pinned identity is the one a later look finds, that `st_ctime` is the *only* field dropped and only across the two APIs, and the end-to-end statement that a scan-only adopt finds the pages that are there.

Ten adjacent adoption and census suites: **15 failures before, the same 15 after** — the branch-only set is empty.

`uvx ruff check . --select F` passes.

### DIFF
--- a/src/exomem/semantic_census.py
+++ b/src/exomem/semantic_census.py
@@ -43,6 +43,27 @@ def _is_reparse(info: os.stat_result) -> bool:
     return bool(getattr(info, "st_file_attributes", 0) & marker)
 
 
+def _identity_bearing_stat(entry: os.DirEntry[str]) -> os.stat_result:
+    """A stat result that actually carries this entry's identity.
+
+    `os.DirEntry.stat()` answers from the directory-enumeration record. On
+    Windows that record holds no volume serial and no file index, so `st_dev`
+    and `st_ino` both come back 0 -- while every re-check in this module reads
+    them through `lstat` or `fstat`, which open the object and get the real
+    pair. Pinning an entry to the enumeration record there pins it to an
+    identity nothing can match: the census counted every subdirectory as having
+    changed under it, descended into none of them, and reported an empty vault
+    to a scan-only adopt.
+
+    So take one real look, at the two places an identity is about to be pinned.
+    POSIX fills both fields during enumeration, so the record is returned
+    unchanged and no extra call is made there.
+    """
+    if os.name != "nt":  # pragma: no cover - the Windows branch is the fix
+        return entry.stat(follow_symlinks=False)
+    return os.lstat(entry.path)
+
+
 def _same_identity(left: os.stat_result, right: os.stat_result) -> bool:
     return (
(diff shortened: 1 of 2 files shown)


Artexis10:
Superseded by #642, which landed the same two repairs independently: `_comparable_identity_stat` for the enumeration record that carries no file id on Windows, and `_same_observed_state` for the cross-stat-API `st_ctime` comparison. `tests/test_adopt.py` is 46/46 on `main` at 42c94959, so there is nothing left here to land.

PR #686, opened 2026-08-19
fix(hosted): make the promotion harness survive its own window

Artexis10:
Three defects in `scripts/reviewer_bootstrap.py`, each of which only surfaces *after* the invite is spent and the ≤30 minute authority clock has started — the one place where no retry can recover.

## 1. `run` could not start at all

`main()` passed `args.openai_redirect` into `run()`, but no `add_argument` ever declared it:

```
AttributeError: 'Namespace' object has no attribute 'openai_redirect'
```

The flag was documented in two places — `--openai-connector`'s help text and `chatgpt_cimd_identity`'s docstring, which tells the operator to reach for it when the publisher's bot protection answers a residential address with 403 where it answers the control plane with 200. The operator would have been told to use an option that did not exist, in the one situation where the alternative had already failed.

## 2. The OpenAI sibling stage could never be created

Every contract candidate is created with `openai_package_lock` NULL — `storeExomemAgentContractCandidate` sets it so deliberately, because the OpenAI artifact is not part of the checked Exomem release. `createStagedClientRelease` for platform `openai` joins on:

```sql
AND candidate.openai_package_lock->>'artifact_sha256' = <packageSha256>
```

A null lock matches nothing. `run` builds the Claude sibling first, so this lands mid-window. `prepare` now calls `attach-openai-locks` as its first step, signing the payload with the control plane's canonical encoding.

## 3. Repo locks silently drifting from the candidate

The locks come from `--repo`, and must be the ones the candidate was cut from — not whatever HEAD generates. Exomem #608 (`feat(sources): open the source taxonomy`) moved `schema_contract_sha256` and `compatibility_sha256` on 2026-08-18, the day after candidate `eb88eedb` was cut, without moving the packaged artifact digest. Every server-side join on those digests then fails: `create-stage` with a bare 500, `attach-openai-locks` with a silent `false`.

Preflight now compares all three digests against the candidate and names the field that moved.

## Verification

Against production, read-only:

```
$ preflight --repo .                      # this repo's HEAD (0.56.0 locks)
  ok    candidate eb88eedb: pending 0.54.1 routable=0
  FAIL  repo locks match candidate
          schema contract: repo 30c65de187984940 != candidate 471cd6bf03cacfe0
          compatibility:   repo 9e028c9e2001378a != candidate 54c7a376d5bca6ec

$ preflight --repo <worktree at the deployed revision>
  ok    repo locks match candidate
```

The deployed cell independently corroborates which side is correct: `exomem_routable_cell_contracts` records the last provisioned cell reporting `source_release=0.54.1`, `contract_digest=471cd6bf…`, `compatibility_digest=54c7a376…` — matching the candidate, not HEAD.

The canonical JSON encoder was checked byte for byte against the control plane's own `canonical()` over the real 1242-byte lock payload.

14 tests in `tests/test_reviewer_bootstrap_cli.py`, all passing; `ruff check` and `ruff format --check` clean. The CLI-surface tests read the parser and dispatcher with `ast` rather than a hand-kept list, so a future option consumed but not declared fails in CI instead of live. Mutation-checked: re-removing `--openai-redirect` fails five of them.


Artexis10:
Diagnosed the `tests (windows-latest, py3.13, shard 4/4)` failure — it's a real cross-platform bug in the new test, not a flake. Not touching your branch; here's the fix.

```
>  assert (tmp_path / "bootstrap-outcome-final.json").stat().st_mode & 0o777 == 0o600
E  AssertionError: assert (33206 & 511) == 384
```

`33206 & 0o777` is `0o666`. Windows **synthesizes** mode bits — a file reports `0o666` and a directory `0o777` whatever `chmod` was asked for. The assertion isn't reading a permission there, it's reading a placeholder.

## Why the existing 0o600 assertions on `main` don't hit this

`tests/test_hosted_security.py:270` makes the same assertion and is green on Windows, which looks contradictory until you see what follows it:

```python
assert database.stat().st_mode & 0o777 == 0o600
assert database.stat().st_uid == os.geteuid()      # <-- AttributeError on Windows
```

`os.geteuid()` doesn't exist on Windows, and `tests/conftest.py` has a `pytest_runtest_makereport` hook that converts exactly that into a skip ("the hosted cell runtime's POSIX APIs do not exist on Windows"). So the whole test is skipped before the mode assertion can be judged.

Your new test asserts the mode bits **without** a POSIX API call after it, so it fails with a plain `AssertionError` — which the hook deliberately will not absorb, since matching on assertion text would let it swallow real defects.

## Suggested fix

`tests/benchmark_capabilities.py` already exposes the predicate, with this exact case in its docstring:

```python
from benchmark_capabilities import has_posix_file_modes

if has_posix_file_modes():
    assert (tmp_path / "bootstrap-outcome-final.json").stat().st_mode & 0o777 == 0o600
```

or `@pytest.mark.skipif(not has_posix_file_modes(), reason="Windows synthesizes st_mode")` if the whole test is about the permission.

The Linux lane keeps gating it, which is where the permission is an access-control fact rather than a placeholder.

---

Separately: this branch has been updated from `main`, so it now carries #693's fix for the `test_background_media_hashing...` flake. The mode-bits failure above is the only real one left on it.

Issue #283, opened 2026-07-20
Read path: warm hybrid search is ~7s at 2.4k pages (vector 3-6s, BM25 3.2s, query cache never hits)


Artexis10:
Measured 2026-07-20 on the production writer (desktop, 0.26.0, compute mode "performance", 2,444 KB pages) via `ask_memory(include_timings=true)`, same query twice back-to-back:

| Stage | Call 1 | Call 2 (immediately after) |
|---|---|---|
| total | 7,024 ms | 6,818 ms |
| vector | 6,286 ms | 3,141 ms |
| bm25 | 37 ms | 3,189 ms |
| keyword | 245 ms | 181 ms |
| rerank (auto) | 230 ms | 89 ms |
| cache | enabled, **miss** | enabled, **miss** (identical query!) |

Two problems:

1. **Warm hybrid search is ~27x over the ~256 ms worst-case the 2026-07-05 perf arc achieved** (PRs #124/#125/#127-#129). The corpus roughly doubled in July (2,444 pages, .embeddings.sqlite 452 MB, .lexical.sqlite 642 MB) — this looks like the read-side sibling of the write-path O(corpus) regression fixed in #282: nothing broke, the corpus outgrew fixed-cost assumptions. Vector and BM25 lanes trade 3s+ costs between calls, suggesting per-call index (re)load rather than a resident index.
2. **The query cache never hits** — identical query, same session, seconds apart, `cache: {enabled: true, hit: false}` both times. Either the key includes something that varies per call or freshness invalidation churns on every request. A working query cache would make repeated agent lookups near-free regardless of lane cost.

Suggested instrumentation-first approach, mirroring #282: log what the vector/bm25 lanes spend their time on (model load? sqlite open? full-scan?) and why the cache key misses, then fix the dominant term. The idle reaper (15 min, all modes — `release_when_idle` unconditional) likely interacts: a lane that lazily reloads a 452 MB sidecar after eviction will pay seconds per query in exactly the pattern above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Artexis10:
## Re-measured on 0.55.0 (2026-08-19), same method as the original

Identical query twice back-to-back via `ask_memory(include_timings=true)`, live vault.

| Stage | 2026-07-20 (0.26.0) call 1 | 2026-07-20 call 2 | **2026-08-19 (0.55.0) call 1** | **2026-08-19 call 2** |
|---|---|---|---|---|
[...]

PR #695, opened 2026-08-20
perf(find): let a repeated query survive a write that cannot have changed its answer

Artexis10:
The read cache does not work on the real vault. From #283, the *same query,
twice, seconds apart, same session*: `cache: {enabled: true, hit: false}` both
times, 7,024 ms then 6,818 ms. Measured again on 2026-08-20 at ~2,400 pages:
`cache.hit: false` on both of two back-to-back identical calls.

## What actually varied

The brief required this be diagnosed before anything was changed, because a key
that is non-deterministic across calls would be a much simpler and much worse
bug than a key that is merely too broad.

It is not non-determinism. With live-seeded freshness, **two identical no-write
keys compare equal**. One unrelated page write changed exactly **key elements 2
and 3 — the `kb` and `vault` recall projections**. Nothing else moved.

So `_freshness_key` was working as designed, and the design was the problem: a
one-page edit anywhere invalidates the cached answer to every unrelated query.
On a vault written to constantly from inside agent sessions, that is a cache
that never hits.

## The change

Cached entries now carry a durable recall checkpoint alongside them, and a hit
is allowed only when nothing that could alter *this* answer has happened.

**Everything load-bearing in the old key stays.** The sidecar
`(epoch, generation, instance)` write tokens are untouched — the docstring is
explicit that keying on file mtime instead produces both spurious misses and
stale hits, and that stays true. The access-policy fingerprint remains
all-or-nothing, because a policy change can flip admission for every path in
the vault. The daily expiry, the semantic-language and relation registries, and
the graph, embedding and lexical tokens all still gate as before.

What is added is a bounded delta check, and only for plain unfiltered keyword
queries. A cached result is reused only when every represented recall scope
returns a **complete** delta, every global correctness fence is unchanged, and
every changed page is proven unable to affect the answer.

It misses conservatively, on all of: deletions, catalog-only movement, external
pending writes, more than 64 delta paths, missing or unstable signatures,
changes to a page that was in the cached hit, a page that newly matches,
`parent_media`, and scene-frame changes.

**Hybrid and vector queries, and every filtered query, stay on full-key
invalidation**, because a textually unrelated document can still move semantic
or corpus-wide ranking. This optimisation deliberately does not reach them.

## Measured, 2,400 pages, freshness seeded for both scopes

`documents=2400, tokenized_documents=2400` asserted before any timing was
trusted.

| | before | after |
|---|---:|---:|
| repeated identical query | 1.13 ms hit | 1.02 ms hit |
| **after one unrelated write** | **5,391.6 ms miss** | **12.1 ms validated hit** |

The first-request numbers (6,877 ms vs 4,391 ms) include page-cache warming and
are not the comparison that matters; the post-write row is.

## Tests

`tests/test_query_cache_scope.py`, 7 cases: deterministic and repeated hits, an
unrelated write, a *matching* write (the one that makes the optimisation safe
rather than fast), access-policy invalidation, an incomplete delta, and
scene-frame parent invalidation.

## Verification

Draft so CI runs concurrently. Locally: the new file 7 passed;
`tests/test_retrieval_golden.py` 5 passed after `uv sync --extra embeddings`,
with the lean environment restored afterwards and no lockfile change; `ruff
--select F,E9` clean; `validate-public-artifacts --repository` clean; `git diff
--check` clean.

Reported rather than claimed: the full suite exceeded a one-hour bound while
four other workers were running, and `tests/test_latency_gate.py` hit the
repository's 60-second timeout inside unchanged `recall_policy` /
`find_policy` code — note that fixture sets `EXOMEM_FIND_CACHE_SIZE=0`, so the
path added here is disabled there. CI is authoritative for both.

Refs #283.

### DIFF
--- a/src/exomem/find.py
+++ b/src/exomem/find.py
@@ -172,6 +172,9 @@ def auto_rerank_allowed_by_policy() -> bool:
 # --------------------------------------------------------------------------- #
 _FIND_CACHE: OrderedDict[tuple, list[Hit]] = OrderedDict()
 _FIND_CACHE_LOCK = threading.Lock()
+_FIND_CACHE_CHECKPOINTS: dict[
+    tuple, tuple[tuple[str, freshness.RecallFreshnessCheckpoint], ...]
+] = {}
 _DEFAULT_FIND_CACHE_SIZE = 32
 # A request gets a new FreshnessSnapshot, but a live recall projection changes
 # only when its exact checkpoint moves.  Rebuilding the vault-relative allowset
@@ -185,6 +188,7 @@ _RECALL_PATH_CACHE_LOCK = threading.Lock()
 _RECALL_PATH_CACHE_SIZE = 32
 MAX_RERANK_CANDIDATES = 300
 _FOREGROUND_LEXICAL_REPAIR_PAGE_CAP = 64
+_FIND_CACHE_DELTA_PATH_CAP = 64
 
 
 def _bounded_lexical_repair_allowed(freshness_key: tuple | None) -> bool:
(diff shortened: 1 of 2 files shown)


Artexis10:
Reviewed the correctness argument rather than taking the tests' word for it,
since a false hit here is silently stale recall — the worst failure this product
has.

**The match predicate is exact, not approximate.** The reuse check rejects a
changed page with `page.parent_media or _make_excerpt(page, query_norm) is not
None`. I went looking for a gap between that and the real keyword gate, because
an approximation would let a newly-matching page slip past. There is none:
`find.py:3749`, inside the reference keyword scan itself, is
`if _make_excerpt(page, query_norm) is None: continue`. Same function, same
argument. `make_excerpt` returns `None` iff some query token is absent from both
`title_norm` and `body_norm`, which is the all-tokens-present contract the
docstring claims.

Consequences worth stating, because they are what make the narrow scope safe:

- a matching changed page always misses, whether or not it was in the cached
  answer — so a page that newly outranks the `limit` boundary cannot be missed
- a changed page already in the cached hits misses unconditionally
- a non-matching changed page cannot enter the result at all, so its `updated`
  moving is genuinely irrelevant to the `updated`-desc ordering

**The fences that must not be scoped are not scoped.** `_freshness_correctness_fences`
strips exactly `kb`, `vault` and `lexical`; the date roll, the access-policy
fingerprint, the semantic-language and relation registries, and the semantic and
graph sidecar write tokens all remain unconditional. The sidecar
`(epoch, generation, instance)` keying is untouched, which matters — the
docstring records that keying on file mtime instead produces both spurious
misses and stale hits, and that was a hard-won fix.

**The TOCTOU handling is right.** `stat_signature` is compared against the
delta's target both before and after `_CACHE.get`, `external_pending` is checked
at entry and again at the end, and every advanced checkpoint is re-read and
compared before the hit is allowed. A file changing mid-validation falls back
rather than being served.

One deliberate non-optimisation, noting it so it does not read as a bug: an
empty delta returns `False`. Reaching this function at all means the key already
differs, so an empty delta is a contradiction and falling back is the safe
reading of it.

No changes requested. Holding the merge only on CI.

PR #702, opened 2026-08-20
fix(ci): give the contention tests a hold/observe shape instead of tight literals

Artexis10:
Two sweeps of the same underlying defect: a wall-clock number that exists so a
hang fails instead of blocking forever, written tightly enough that it also
claims a contended CI runner is fast.

## Part 1 — the last four sub-second assertions

`tests/` had four wall-clock bounds under one second. They are **not all the
same defect**, so they do not all get the same fix.

**Two pure deadlock valves — widened.** `test_client_artifacts.py`. Each hands
the code a `0.01s` deadline then blocks the underlying call indefinitely. If the
deadline never fires the call never returns, the `release.set()` in `finally`
never runs, and **the test hangs** rather than overrunning. Any bound catches
that, so it may as well be one a loaded runner cannot reach.

**One redundant proxy — demoted to a valve.** `test_continuation_checkpoint.py`
already asserts `inspected <= 32` against a 300,000-entry listing, which proves
the enumeration cap directly. `elapsed < 0.1` restated it more weakly and was the
only one of the two a loaded runner could fail.

**One genuinely discriminating — kept, given headroom properly.**
`test_mutation_lock.py`: the bound proves the contender gave up on its *own*
`0.05s` timeout rather than waiting out the holder. Relaxing it would make it
vacuous, because the holder only held 2s. So the bound is not relaxed — the
**gap** is widened: the holder now holds 60s against a 15s observation.

## Part 2 — finishing the graph availability suite

`#689` introduced `_HOLD_SECONDS`/`_OBSERVE_SECONDS`/`_ADMIT_SECONDS` in
`test_graph_rebuild_availability.py` but converted only the tests failing that
day. The 29 literals left behind kept the original shape — a 2s hold guarding a
1s observation.

**That came back today.** On PR #694, a Windows shard failed
`test_legacy_refresh_releases_canonical_boundary_before_missing_sidecar_rebuild`
on `assert rebuild_started.wait(1)`: the runner had not yet scheduled the rebuild
thread. Nothing was wrong with the product.

All 29 are converted on one rule:

| shape | meaning |
|---|---|
| `release*.wait(...)` | a **HOLD** — parks a lock or in-flight rebuild while the test observes an ordering. Must outlast the observation or the ordering proves nothing. |
| other `.wait(...)` | an **OBSERVATION** — how long the test waits for something to become true. Generous, because the hold is far larger, so a wrong ordering still fails. |
| `join(timeout=...)` | a **DEADLOCK VALVE** — the `is_alive()` assertion after it is the real check. |

The constants also move to the top of the module, where someone adding a test
will look, rather than 2,200 lines in.

## Verification

- `test_graph_rebuild_availability.py`: 107 passed, 4 skipped.
- `test_mutation_lock.py` + `test_client_artifacts.py`: 108 passed.
- `test_continuation_checkpoint.py`: 147 passed, 20 skipped.
- `uvx ruff check --select F,E9` clean.
- No sub-second wall-clock assertion and no literal `.wait()`/`join(timeout=)`
  remains in either swept file.

## Not claimed

A sweep of the whole suite finds **~250** literal sub-5s waits across 35 files.
Most are not this defect — an event set synchronously never needs a generous
window. This PR fixes the two files with a demonstrated CI failure and does not
touch the rest; blanket-widening would destroy real signal in tests where the
bound is the assertion.


Artexis10:
Extended to the two files that failed CI on this same defect after the first pass: `test_mutation_lock.py` and `test_epistemic_graph_freshness.py`.

**Why they survived the first pass:** the sweep matched `.wait(2)` but not `.wait(2.0)`. Every decimal literal went straight through it.

54 more sites converted, classified by the name being waited on.

## The part worth reading — three sites deliberately left tight

`assert not X.wait(0.05)` is a **negative** observation: it proves something has *not* happened yet. Widening one does not merely slow the test, it **changes the scenario**, because the product's own acquire timeout runs during that same window.

At 15s, the holder in `test_probe_cleanup_cannot_delete_a_new_holders_metadata` gave up with `MUTATION_BUSY` *inside* the settle period and never entered afterwards — so the positive assertion that followed failed:

```
>       assert holder_entered.wait(15.0)
E       assert False
    exomem.cli_ops.OpError: {"code":"MUTATION_BUSY", ... "wait_ms":2258.45}
```

Blanket-converting them broke three tests locally before this reached CI.

The asymmetry is the whole point: a slow runner can only make a *negative* wait pass **vacuously**, never fail. A *positive* wait on a slow runner fails a correct product. That is why negative waits are safe left tight where positive ones are not — and they stay as literals, deliberately, so a future sweep does not carry them along, with a comment saying so.

## Verification

`test_mutation_lock.py` + `test_epistemic_graph_freshness.py`: **133 passed, 1 skipped**, in 44s — down from 142s, because the broken conversion had tests sitting out full 15s settle periods.

PR #705, opened 2026-08-20
fix(codex): give lane workers full access so briefs stop asking the impossible

Artexis10:
**The Codex worker protocol has been asking for deliverables the sandbox made impossible.**

A linked worktree's gitdir lives under the primary checkout's `.git/worktrees/`,
which is outside a `workspace-write` sandbox. So a worker could not create
`index.lock` and could not commit its own work. The same sandbox blocked
`~/.cache/uv` and every PyPI route, so an acceptance command naming `uvx` could
never pass either.

The brief template and `CLAUDE.md` asked for all three anyway. A worker that
skips a structurally impossible step reads afterwards exactly like one that
ignored its brief — which is how the `export-privacy` lane read until the
sandbox limits were checked against the KB note that already documented them.

## Change

`codex exec` now runs `-s danger-full-access`.

## The cost, stated rather than implied

The sandbox no longer confines a worker to its worktree.
`require_linked_worktree` still decides where a lane may *start*, but nothing
stops a worker reaching the primary checkout mid-run — and this repo has
already had one mid-edit collision. The brief's scope allowlist and
`codex_task.sh verify` are what catch it, and both act after the fact.

`CLAUDE.md` claimed `codex exec` runs "`workspace-write`, sandboxed to the
worktree — never on the primary checkout (the runner enforces this)". The
second half of that is now false, so it is rewritten rather than left standing.

## Alternative considered

`workspace-write` with `writable_roots` pointing at the gitdir would restore
in-lane commits without opening the filesystem, but leaves the uv-cache and
PyPI limits in place. Full access was chosen deliberately; the narrower option
remains available if the isolation loss proves to matter.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -117,10 +117,15 @@ Lane mechanics: one lane = one sibling worktree (`../exomem-<lane>`, branch
 `codex/<lane>`, from `origin/main`) = one self-contained `.task/TASK.md` brief
 (`codex_task.sh template`) naming the OpenSpec artifacts as source of truth,
 a scope allowlist, and exact acceptance commands. `codex exec` runs
-`workspace-write`, sandboxed to the worktree — never on the primary checkout
-(the runner enforces this). Results come back as commits on the lane branch
-plus `.task/RESULT.md`; briefs live under `.task/` (git-excluded, never
-committed). Before merging, `scripts/codex_task.sh verify <worktree>` must
+`danger-full-access`, started only in a linked worktree (the runner enforces
+where a lane *starts*, not where it stays). Full access is deliberate:
+`workspace-write` put a linked worktree's gitdir out of reach, so workers could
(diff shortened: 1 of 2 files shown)


Artexis10:
Superseded by #700, which landed the same decision with more mechanism around it: a `CODEX_SANDBOX` override rather than a hardcoded flag, a deterministic preflight that proves the environment can run a test before a lane is handed an hour of work, the platform pytest-ignore policy mirrored from cross-platform.yml, and `tests/test_codex_lane_contract.py` pinning all of it.

Your rationale was better than mine and it is not lost — #700 now carries the two causes it was missing (a linked worktree's gitdir sits outside the sandbox, so a worker could not create index.lock or commit; `~/.cache/uv` and PyPI were blocked, so no brief could name `uvx`), plus your honest statement of the cost: full access does not confine a worker to its worktree, and the scope allowlist and `verify` both run after the fact.

Thanks — two sessions reaching the same fix independently is a good signal it was the right one.

PR #708, opened 2026-08-20
perf(recall): build the cold resolver from the lexical sidecar instead of the vault

### LINKED ISSUE #676: Read path: a cold recall resolver rebuilds the whole vault on the reader's thread — measured 30.7s of a 34s call
Measured 2026-08-19 on the production writer (0.55.0, compute mode "performance") via `ask_memory(include_timings=true)` on the live vault. Three calls, same warm process:

| Call | total | graph stage | backend |
|---|---|---|---|
| 1 — query A | 6,118 ms | `graph: 216 ms` (`graph.sidecar: 184 ms`) | `epistemic_graph` |
| 2 — query A repeated | **71 ms** | – | query-cache hit |
| 3 — **query B** | **34,052 ms** | **`graph: 30,787 ms`** — `graph.resolver: 30,763 ms` | **`wikilink_fallback`** |

Call 3's other lanes are all fast: `vector: 317 ms`, `bm25: 24 ms`, `keyword: 121 ms`, `fusion: 74 ms`, `freshness: 1,279 ms`. **90 % of a 34-second read is one stage**, and that stage is not retrieval — it is building a resolver.

## What `graph.resolver` is

`find_candidates.py:625` — the `wikilink_fallback` branch, taken when the epistemic-graph sidecar is not usable for the request. It calls `get_query_resolver(...)` → `find.recall_resolver_snapshot` (`find.py:3678`). On a cache miss that body is:

```python
entries = []
for path in walk_vault_md(root):
    if not recall_policy.is_recall_candidate(root, path):
        continue
    page = _CACHE.get(path, root)           # parse
    if page is not None:
        entries.append((page.rel_path, page.title))
resolver = WikilinkResolver.from_entries(root, entries)
```

A full walk of every markdown file plus a parse per page — O(vault), synchronously, **on the reader's thread**, at ~2.4k pages.

## It is the COLD build that is unbounded, not the warm path

The warm path is fine and deliberately built: `on_resolver_files_changed` (`find.py:3914`) patches `_RECALL_RESOLVER_CACHE` incrementally from the exact event checkpoint, with path guards, a re-check before publication, and a freshness restamp so the next query hits instead of rebuilding. It is called from the watcher, `index_sync`, `note`, `link` and `replace`. A vault that is merely being *edited* does not pay this.

What has no bound is the **miss**. Two routine things empty the cache:

1. **The idle reaper.** `unload_ram_caches()` (`find.py:4065`) clears `_RECALL_RESOLVER_CACHE` outright, and the measured policy has `release_when_idle: true`.
2. **Never having been built in this process for this lane.** Call 1 took the sidecar branch, where `get_query_resolver` is reached only `if legacy_seeds:` — so a process can serve graph-lane recalls for a long time without ever building the recall resolver, and then the first request that falls back pays the entire cost. That is what calls 1 and 3 show: same process, one minute apart, no idle window in between.

`warmup.py:112` warms it once at startup, which does not survive (1) and does not help a lane that first needs it later.

So the shape is: **an unbounded O(vault) build, on a miss, paid synchronously by whoever asks first — and that is a reader.** This is the read-side twin of #539, where populate-on-miss made a *writer* pay a full corpus build. There the fix was single-flight plus a memo; here the answer is probably different (a reader cannot usefully be told "no"), but the diagnosis is the same one.

## Why the fallback is reached at all is a second question

Call 1 used the `epistemic_graph` backend; call 3 fell back. Graph availability is a single vault-global equality (`epistemic_graph.py:1133`, `stored_projection == current_identity`), so one write anywhere makes the whole graph unavailable and every recall after it takes the fallback until a rebuild lands.

The two defects **compose**: the fence flips often, and the path behind it is the most expensive operation on the read path.

## Suggested order

1. **Bound the cold build.** Options worth weighing: single-flight it so a burst of readers builds once rather than N times; build it incrementally from the checkpoint the incremental patch already understands rather than from a bare walk; or keep the reader on a degraded-but-correct lane while the build runs in the background, the way the deferred-index family already does. What must not survive is "the first reader after an idle window pays 30 seconds".
2. **Reconsider the reaper's reach.** `release_when_idle` unloading a structure that costs 30 s to rebuild is a bad trade unless memory is genuinely scarce.
3. **Only then** revisit how often the fallback is taken (the availability fence). If the fallback is cheap, the fence matters much less — the same argument the incremental-graph plan makes for deferring its Phase 3.

## Relationship to #283

#283 measured this read path in July and named vector/BM25/query-cache as the costs. Re-measured today, [the query cache is fixed and the lane costs are down 4-5x](https://github.com/Artexis10/exomem/issues/283#issuecomment-5342852677). This resolver build is larger than everything #283 names put together, and it is absent from that report because both of its calls took the sidecar path.

---

*Correction: the first version of this issue claimed `_RECALL_RESOLVER_CACHE` had no incremental warm path at all. That was wrong — it was read off a truncated grep. The patch path exists and is thorough; the defect is the unbounded cold build behind it. The measurement is unchanged.*

Artexis10:
Closes #676 — a cold recall resolver rebuilt the whole vault on the reader's thread, measured at 30.7s of a 34s call.

Implemented by a Codex worker lane (`codex/resolver-titles`) under the protocol in CLAUDE.md; reviewed and pushed here.

## Change

The resolver needs `(rel_path, title)` for every recall-admitted page. It got them by walking the vault and parsing every page — on the reader's thread, on a cold cache. The lexical sidecar already stores a row per page, so the titles can come from one SQLite query instead.

- Lexical sidecar schema 6 → 7, adding nullable `pages.title`, persisting the parser-produced `page.title` unchanged under the canonical `page.rel_path` key.
- `find.recall_resolver_snapshot` reads `(path, title)` in a single query and **retains the walk-and-parse implementation as its fallback**.
- No handwritten migration: `_schema_is_current()` rejects a non-current version and the ordinary repair path rebuilds the disposable sidecar.

## Measured — 2,400 pages

| path | cold resolver build | `find._CACHE.get` | retained resolver bytes |
| --- | ---: | ---: | ---: |
| forced legacy walk | 4.991 s | 2,400 | 1,461,009 |
| sidecar SQL | **0.097 s** | **0** | 1,460,993 |

**51× faster, 2,400 page reads down to zero, and the same ~1.39 MiB retained** — this trades no memory for the time.

## Correctness

- `pages.title` is the exact parser value including case and Unicode; SQL `NULL` stays Python `None`.
- Scope flags are honoured only once the stored projected checkpoint equals the current checkpoint, which includes policy and access identity. Both the sidecar walk and the insert path apply `is_recall_candidate`, and the regression test compares every `in_vault` resolver row against a fresh policy walk.
- Missing, stale, wrong-schema, or SQLite-failed sidecar → `None`, and `find` runs the original complete walk. That fallback deliberately does **not** kick off an async rebuild, so a read-only resolver call never creates temporary SQLite files.

## Verification

`tests/test_resolver_from_sidecar.py` (8 new tests): title parity and exact casing/Unicode, no parser calls on the fast path, `NULL` title, absent/stale/wrong-schema fallback, policy change, unknown sidecar page, and no background repair from a fallback.

- `tests/test_lexstore_atomic_rebuild.py` — 40 passed
- `ruff check . --select F,E9` — clean
- `tests/test_retrieval_golden.py` — skipped, `sentence_transformers` absent in that environment

The lane's full local suite did **not** reach a terminal state on that host and the worker correctly declined to claim it green. The merge gate is running locally; CI here is the real check.

### DIFF
--- a/src/exomem/find.py
+++ b/src/exomem/find.py
@@ -4021,6 +4021,7 @@ def recall_resolver_snapshot(vault_root: Path, freshness: tuple | None = None):
     an ordinary note in the graph lane.
     """
     from . import freshness as freshness_module
+    from . import lexstore
     from .vault import WikilinkResolver, walk_vault_md
 
     root = Path(vault_root)
@@ -4088,13 +4089,17 @@ def recall_resolver_snapshot(vault_root: Path, freshness: tuple | None = None):
     # would then elect itself leader and rebuild -- reintroducing the stampede
     # this exists to prevent, just narrowed to a race window.
     try:
-        entries: list[tuple[str, str | None]] = []
-        for path in walk_vault_md(root):
-            if not recall_policy.is_recall_candidate(root, path):
-                continue
-            page = _CACHE.get(path, root)
-            if page is not None:
-                entries.append((page.rel_path, page.title))
+        entries = lexstore.get_store(root).recall_resolver_entries(
+            "vault", checkpoint.triple if checkpoint is not None else None
+        )
+        if entries is None:
+            entries = []
+            for path in walk_vault_md(root):
+                if not recall_policy.is_recall_candidate(root, path):
+                    continue
+                page = _CACHE.get(path, root)
+                if page is not None:
+                    entries.append((page.rel_path, page.title))
         resolver = WikilinkResolver.from_entries(root, entries)
         with _RESOLVER_LOCK:
             # The supplied freshness key names this immutable resolver snapshot.
--- a/src/exomem/lexstore.py
+++ b/src/exomem/lexstore.py
@@ -71,7 +71,7 @@ from .kbdir import kb_dirname
 log = logging.getLogger(__name__)
 
 _NAV_BASENAMES = frozenset({"index.md", "log.md"})
-SCHEMA_VERSION = 6
+SCHEMA_VERSION = 7
 CATALOG_FOREGROUND_DELTA_CAP = 32
 
 # Publication-barrier timeouts. Every LIVE-sidecar mutation and journal-mode
@@ -1406,19 +1406,16 @@ class LexicalStore:
             except OSError:
                 pass
 
-    # The current normal-table shape sentinel: an older (v4) `pages` table lacks
-    # this column, so its presence distinguishes a mutation-safe current shape
-    # from a legacy shape a new-column index/INSERT would fault against.
-    _CURRENT_PAGES_COLUMN = "emitted_parent_path"
+    # The current normal-table shape sentinels. An older `pages` table lacks
+    # one or more of these, so it must be rebuilt before a new-column INSERT.
+    _CURRENT_PAGES_COLUMNS = frozenset({"emitted_parent_path", "title"})
 
     @staticmethod
-    def _pages_has_emitted_parent(conn: sqlite3.Connection) -> bool:
-        """Read-only: does the `pages` table carry the current emitted-parent
-        column? False when the table is absent or is an old (v4) shape."""
-        return any(
-            row[1] == LexicalStore._CURRENT_PAGES_COLUMN
-            for row in conn.execute("PRAGMA table_info(pages)")
-        )
+    def _pages_has_current_columns(conn: sqlite3.Connection) -> bool:
+        """Read-only: does `pages` carry every current schema sentinel?"""
+        return LexicalStore._CURRENT_PAGES_COLUMNS <= {
+            row[1] for row in conn.execute("PRAGMA table_info(pages)")
+        }
 
     def _schema_is_current(self, conn: sqlite3.Connection) -> bool:
         """Read-only probe: is the catalog schema at the current version AND the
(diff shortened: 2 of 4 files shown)


Artexis10:
The local merge gate's failure was environmental, not a regression — recording the control so nobody re-investigates it.

```
lane worktree (this PR):  test_thousand_planning_items_... -> Timeout after 61s
origin/main (control):    test_thousand_planning_items_... -> Timeout after 63s
```

Same test, same timeout, with and without this change. The stack bottoms out in `_open_windows_path_descriptor` / `os.stat` — Windows file-handle contention on a box that had 34 concurrent pytest processes at the time, 17 of them orphaned from earlier runs. Not a memory problem (those orphans total 9 MB); a handle-pressure one.

CI, which is the cleaner signal, reported exactly one failure on **both** the Linux and Windows shards: `test_hierarchy_parser_and_sidecar_versions_are_incremented`. That one was real, and it is fixed in 7f25b524.

PR #725, opened 2026-08-21
feat(memory): surface due-state counts on writes, recall and bootstrap

Artexis10:
## Why

The epistemic loop's representation layer shipped (predictions with `check_by`, verdicts, experiment outcomes), and PR #555 gave two of its obligations a queue — but nothing *delivers* any of it. Review queues run only when an agent remembers to call them, hooks exist only on the CLI clients, and hookless surfaces rely on prose that decays. The 2026-08-15 no-nudge audit measured the result: the substrate detects, the conversation never hears about it, and mid-conversation accumulation is invisible until an expert user intervenes.

This is the keystone slice of that programme (S1 of `add-due-state-consumers-and-carriers`): one maintained due-state projection, served on the three channels every client has.

## Reconciliation with #555 (the "coordination decision" its PR body asked for)

#555 implemented `prediction_window` and `unfinished_experiments` under `add-prediction-window-review` and `close-experiment-lifecycle`, with exactly the unit-local predicate this change's design demanded. This change therefore **consumes** those two categories and adds only the remaining two; its `attention-queue` delta was rewritten so that three active changes no longer add the same requirement (the archive-sync collision #555 flagged). The two #555 changes are untouched.

## What this does

- **Two new audit categories.** `question_aging` (a governed question unit on an active page with no answering structure after a provisional 30-day threshold; `info`, "a candidate, never a defect"; registered but opt-in because the threshold is system-invented) and `supersession_integrity` (dangling `superseded_by` targets and multi-headed chains; `warn`; joins the default attention union — the pointer is human-authored, nothing is inferred).
- **A maintained projection** (`src/exomem/due_state.py`, persisted as `.due-state.json` beside the review state and registered as reserved internal state). It stores pending candidates *with the date they become due*, so a `check_by` passing at midnight surfaces with no write and no audit pass. Per-write deltas are page-local; reconcile heals; missing or corrupt state recomputes at serve time on read surfaces, never inside a mutation.
- **Egress before counting.** Items are filtered through the governance release plane for the requesting audience *before* any count, reference or ordering is computed; a withheld page's obligation is byte-identical to one that does not exist. Fail-closed on a release-plane error.
- **Three carriers, one block.** A bounded advisory `due_state` block (total, per-category counts, ≤5 refs with due dates) on default compact mutating responses — riding the post-commit seam #538 built for `structure_suggestion`, as a sibling field — on recall (delta-only) and in the bootstrap payload, with teaching lines beside `structure_suggestion_handling`. Absent when nothing is due; omitted under legacy detail; no envelope branch key, no tool schema change, packaged tool-surface fingerprint byte-identical.
- **Emission governance** decided at the delivery choke point (not at production, so a legacy-detail write or a terminal validation drop cannot burn the session's emission): first-of-session, change-of-totals, identical totals go quiet, one decision per command invocation, keyed by (session, audience, vault).
- **Triage parity.** One shared composer for review identities (`review_state.component_fingerprint`); every decision recorded through `review_memory`/`triage_memory` fans out to the item's component fingerprints, so a counter goes quiet when its item is dismissed through the ordinary surface. Refs for registered opt-in categories resolve for triage without joining the default union, and without moving the identity of any existing ref.

## Review arc

Fable-delegate discipline: Opus implementer lane → fresh zero-context Opus reviewer (REQUEST_CHANGES: 1 blocking, 3 major, 5 minor, 3 nits) → correction round → targeted recheck → orchestrator acceptance. The blocker was the anti-nag machinery nagging: the projection composed its review fingerprint from a single finding while `attention` composes it from the fused item, so a dismissed item kept counting on every carrier; the original test was tautological (it borrowed the implementation's own fingerprint). Fixed with the shared composer above and cross-boundary tests that dismiss through the shipped command. The correction round also surfaced and fixed an id collision between `unfinished_experiments` items and `relation_debt`-only items, and a cost regression gate 4 caught (the release filter was being built on every `ask_memory` with nothing due).

## Deviations worth knowing

- `COMPACT_BYTE_CEILING` 58,000 → 59,500: headroom was 128 bytes before this change; the three teaching entries cost 1,224. The arithmetic is recorded in the constant.
- `test_default_attention_union_is_pinned` moved because the union widened by `supersession_integrity` — the pin exists to force that argument to be written down, and it is.
- Task 4.4 (live thin-client probes) is deliberately **not** ticked: the portability floor is an assertion about clients, and the evidence gets attached after deploy.
- Bench family f26 is red as scripted against *any* correct implementation because the sample vault owes nothing; green on a seeded variant through the identical CLI envelope. Seeding the journey corpus is a bench-amendment follow-up, not part of this change.

## Verification

Orchestrator-side on the final tree (`dde35e8b`, 5 commits ahead incl. the merge of `origin/main` @ `4bc550ef`):

- Gate 1 — the three new suites (`test_due_state_consumers/projection/carriers.py`): **111 passed**.
- Gate 2 — adjacent contracts (`test_epistemic_review_queues`, `test_attention`, `test_mutation_terminal`, `test_mutation_lock`, `test_bootstrap*`, `test_epistemic_bootstrap_contract`, `test_tool_surface_contract`, `test_plugin_sync`, `test_scaffold_no_leak`, `test_reserved_admin_paths`, `test_structure_promotion`): **472 passed, 11 skipped** (Windows DACL cases).
- Gate 3 — `test_latency_gate` + `test_semantic_write_latency_gate`: **34 passed** (the round-1b short-circuit took `mixed-search` p95 from 10.02 s, failing, back to 5.32 s).
- `uvx ruff check . --select F`: clean; `src/exomem/due_state.py` clean under the repo's full `E,F,I,B,UP,BLE` set.
- `openspec validate add-due-state-consumers-and-carriers --strict`: valid.
- `src/exomem/tool_surface_contract.json`: byte-identical to `origin/main` (`90f6461c…`, 29 tools).
- Lean suite (everything except the two membench modules): see the comment appended below when the run completes.

Red-first evidence, every gate's verbatim output, the mechanism-removal runs (each fix reverted to watch its test go red) and three discarded tautological tests are recorded per round in the lane's `.task/RESULT.md`; the fresh-reviewer recheck closed every finding, then caught two one-line wiring gaps in the emission governor (recall carrier keyed without the vault; bootstrap delivering without marking) which were fixed and pinned through the shipped commands.

Known environmental failures on the development box, reproduced on pristine `origin/main`: the 91 membench tests (need `bun`), `test_bounded_graph_join::test_a_direct_mutation_guard_also_returns_while_its_rebuild_runs` (thread-race assertion, fails 4/5 on main), and `test_governance_overhead::...structure-reduction` (contended-machine timing; its siblings self-skip). CI on a quiesced runner is authoritative for those.


Artexis10:
## Lean-suite verification on the final tree (`dde35e8b`)

Full lean suite (everything except the two membench modules), no early stop, development box:

```
6 failed, 12037 passed, 256 skipped, 6 errors in 2084.26s (0:34:44)
```

Every failure and error was attributed against a pristine `origin/main` worktree (`31b1b6e3`) with `PYTHONPATH` pinned to the baseline `src` (verified: `exomem.due_state` absent there). All are environmental and identical on both trees:

| Module | Lane | Pristine `origin/main` |
|---|---|---|
| `test_lme_runner.py`, `test_lme_runner_protocol.py`, `test_graph_idempotency_crash_matrix.py` | 35 failed, 13 passed, 4 errors | 35 failed, 13 passed, 4 errors |
| `test_report_offline.py`, `test_protocol_schema_conformance.py` | 4 failed, 63 passed, 6 errors | 4 failed, 63 passed, 6 errors |
| `test_bounded_graph_join::test_a_direct_mutation_guard_also_returns_while_its_rebuild_runs` | fails | fails 4/5 runs (thread-race assertion) |
| `test_governance_overhead::...structure-reduction` | fails | fails (contended-machine timing; its siblings self-skip) |

None of those modules touch a file this PR changes. CI on a quiesced runner is authoritative for them.

Focused gates on the same tree: new suites `111 passed`; adjacent contracts `472 passed, 11 skipped`; latency gates `34 passed`; `ruff --select F` clean; `openspec validate --strict` valid; `tool_surface_contract.json` byte-identical to `origin/main`.

PR #755, opened 2026-08-22
feat(review): govern signal families, capture triage metrics and scale the review state

Artexis10:
## Why

S1 (#725) put due-state counts on every write, recall and bootstrap. The moment the substrate starts talking on its own, the failure mode flips from silence to nagging: a family the user never wants to hear about keeps counting, a dismissed item comes back under a fresh reason, a twelve-page adoption could emit twelve blocks, and nothing measures any of it. The no-nudge audit's paired-metric table (nag rate / recovery fraction) needs three captures that do not exist yet: a first-surfaced ledger, closed triage-reason codes, and a manual/automatic origin tag. This is S6 of the programme (`add-nag-governance-and-metrics-capture`): the governance and the measurement that make the S1 channel safe to leave on.

## What this does

- **Family dispositions** (`normal | quiet | off`) as review-state records, set through the existing triage surface: `triage_memory(ref="exomem://review/family/<family>", action=quiet|off|normal, why="<code>: …")`, plus `exomem review quiet|off|normal --reason`. Valid families are the default attention categories, the epistemic review categories and the write-advisory kinds. Effects (design D2): audit always measures; `quiet` leaves the default union, the carriers and the write advisories; `off` is hidden except under `state="all"`; exclusion happens before fusion so a quiet family never lends weight; a disposition never rewrites a decision — an item whose composed reasons change when a family returns to normal resurfaces under the existing new-reason rule with the earlier record still standing.
- **Closed reason vocabulary** (`intentional, false_positive, handled, deferred, too_frequent, unspecified`) parsed from a leading colon token in `why`; unknown tokens are refused, uncoded text is `unspecified`. **Origin** `manual | automatic` on every record.
- **First-surfaced ledger** — the moment an item was first shown to a caller, stamped behind delivery (attention listings and the admitted carrier block), never for withheld or disposition-excluded items, never by audit, never by lookups. Best-effort and lock-scoped; the decision read stays fail-closed (`REVIEW_STATE_INVALID`, as `main`), and a carrier facing an unreadable store serves no block rather than every dismissed item.
- **Emission ledger** in `.due-state.json` (`writes`, `emissions`, `due_total`, `last_digest`) with one definition of `due_total`: the served total at delivery. A `batch_scope` suppresses per-write emission inside multi-write commands (`adopt`, `adoption_studio`, `maintain fix`, `preserve_artifacts`, `process_media`, `backfill-ids`).
- **Review-state schema v2** (`records, dispositions, surfaced, stats`) with v1 migration, retention (snooze 90 d, ledger 400 d), compaction thresholds with a rescan guard, a 64 MiB read limit (measured 41.28 MiB at 50k decisions / 150k ledger rows) and a 4× recovery limit used only by `maintain --reconcile`, which now reports `review_state_compaction` including its failures.
- **`review_memory(mode="dispositions")`** view; disposition-aware write advisories in `corpus_aware`; teaching lines in the bootstrap contract and both review skills; `docs/epistemic-inbox.md` rewritten for the new surface.
- **Bench f23 `dismissal_respect`** gets a real journey driver and a `due_state_counters` projector; the projector maps triage actions onto the review state.

## Deliberate pin move

The `triage_memory` and `review_memory` descriptions say what the surface now does, so the packaged tool-surface digest moves `433e0bb5…` → `30d2e420…`. (Current head `ef95061e`: `89e5a2f4…` — `main` has since moved the pin twice more, most recently in #760, and this branch regenerated its artifacts on each merge; the description change is the same.) No input parameter is added or removed. Regenerated as `docs/remote-quickstart.md` prescribes: `scripts/dump-tool-schemas.py`, the packaged contract, the v1 release-identities fixture, hosted generated locks/directory packets, and the ChatGPT plugin contract's pending digest with `refresh_required: true` (the connector refresh is a post-release step).

## What f23 honestly reports on this runtime

`dismissal_respected_across_passes` passes. `counter_emission_not_repeated_per_write` reports **`unsupported`**, and the journey test pins that verdict rather than a green: no product leaf reaches the write carrier today (measured per leaf — `adopt_vault` 12 governed writes / 0 carrier trips, `adoption_studio apply` 12/0, `maintain fix` 12/0, `preserve_artifacts` 8/0, `process_media` 12/0), so the bulk batch delivers no block, the served denominator is 0, and a pass there would be the assertion agreeing that twelve writes produced no repeat when nothing was produced to repeat. The batch-once requirement is proven where it is decidable: twelve carrier runs inside one scope deliver once, and disabling the scope turns the assertion red (12/12). The spec scenario, design D6 and tasks 5.4 say exactly this. Whether those leaves *should* carry is a product decision this slice leaves open.

## Review arc

Fable-delegate discipline: Opus implementer lane → fresh zero-context Opus reviewer → correction round → targeted recheck → second correction → second recheck → orchestrator acceptance.

- Round 0 review: **REQUEST_CHANGES — 2 BLOCKING, 4 MAJOR, 9 MINOR, 4 NIT.** The blockers were the dispositions view stamping the ledger for every item in the vault, and a corrupt store silently resurrecting every dismissal (the read had been made best-effort along with the write).
- Recheck: **RECHECK: BLOCKED — M3 (journey half inert, tautological response assertion) and m10 (`due_total` had two conflicting definitions)**; everything else CLEAR, eleven new findings.
- Round 2 fixed all eleven (single `due_total` writer, honest f23 verdict, falsifiable negatives, shared fingerprint composer reused instead of re-derived, one `refs_for_paths` per dispositions count — 104 → 2 calls, 127 → 54 ms on a 103-item vault — carrier fail-closed, ref→fingerprint coupling guarded, reconcile failures surfaced).
- Second recheck: **RECHECK: CLEAR — all eleven round-1 blockers verified closed by reproduction, with six new non-blocking findings** (worst: an unbounded SQLite bind list introduced by the `refs_for_paths` hoist — raises past 32,766 paths — and an anti-vacuity gate keyed on a value that persists after the first delivery).
- Round 3 fixed those six (chunked `refs_for_paths`; the anti-vacuity gate now keys on the per-batch emission delta between the journey's prior and later snapshots; doc sentence, import order, a docstring, allowlisted error codes). Orchestrator-verified on the lane tree: the chunk loop in `memory_refs.refs_for_paths` (`REFS_QUERY_CHUNK = 500`, old body moved to `_refs_for_paths_batch` so the direct caller in `commands.py` is covered too), the delta-based verdict in `assertions.py` (writes delta < 2 or emissions delta 0 → `unsupported`; emissions ≥ writes → `fail`), the seven proving tests (7 passed), and both gate-7 probes (journey `PASS` + `UNSUPPORTED`; scope present 12 writes / 1 emission `PASS`, scope disabled 12 / 12 `FAIL`).

## Deviations worth knowing

- `COMPACT_BYTE_CEILING` 59,500 → 60,400 (third raise; headroom is 392 bytes and the warning fires on every run — a compact-bootstrap trim is queued as its own change).
- `test_governance_overhead`'s receipt microgate measures below its own noise floor on this box; not touched here.
- `hosted-alpha-agent-v3` has no committed `openai` render; its `claude.lock.json`/`compatibility.json` were regenerated like the others.

## Verification

Orchestrator-side on the final tree (`37bf42dc`, after merging `origin/main` @ `e4d62e41` and archiving the change):

- Gate 1 — the five S6 suites (`test_review_dispositions`, `test_review_reason_and_origin`, `test_first_surfaced_ledger`, `test_review_state_scaling`, `test_due_state_emission_capture`): **123 passed**.
- Gate 2 — the twenty adjacent suites (attention, review queues, mutation terminal/lock, bootstrap, tool-surface/plugin sync, hosted marketplace release, scaffold no-leak, due-state S1 suites, …): **1008 passed, 16 skipped** (the lane's twenty plus `test_memory_refs`, `test_epistemic_assertions` and `test_hosted_marketplace_release`; skips are the Windows-only branches and `sentence_transformers`).
- Gate 3 — `test_latency_gate` + `test_semantic_write_latency_gate`: **36 passed**.
- `uvx ruff check . --select F`: clean. `openspec validate --all --strict`: 159 passed, 0 failed. `scripts/check_openspec_archive_discipline.py`: `ok` (the change is archived in this PR as `2026-08-22-add-nag-governance-and-metrics-capture`, its three deltas synced into the canonical specs — +9 requirements, 0 modified, 0 removed — per the archive-discipline rule that landed in #742 while this branch was open).
- Lane gate 4 (lean suite, one process, quiet box, pre-merge tree `c360be53`): **12724 passed, 263 skipped in 35m15s, exit 0** — including both tests that failed environmentally in round 1.

Red-first evidence, verbatim gate output per round, and the mechanism-removal runs (every fix reverted to watch its test go red; 11 of 13 removals red in the recheck's own reproduction, 6 more in round 2) are recorded in the lane's `.task/RESULT.md`.

Known environmental failures on the development box (reproduced on pristine `origin/main`): the membench tests (need `bun`), `test_lme_runner*`, `test_graph_idempotency_crash_matrix`, `test_report_offline`, `test_protocol_schema_conformance`, and the thread-race assertion in `test_bounded_graph_join`. CI on a quiesced runner is authoritative for those.

## Post-open correction (2026-08-23)

The required CI gate went red on the stacked PR #758 (py3.11 Linux shard 4/4): `test_review_state_scaling::test_the_stress_gate_holds_at_multi_year_cardinality` measured `apply()` at 1.860 s and then 1.695 s on a rerun against the 1.5 s budget set from a 0.26–0.29 s lane measurement — the hosted runner's fsync-and-replace, not a store change (the test passes locally in 0.30 s; nothing under test moved). Commit `1a5db609` widens the load/apply/compact ceilings to 5.0 s (still an outright failure for the quadratic regression the gate exists to catch; the 2 ms lookup budget is unchanged) and records every timing into the junit via `record_property` so CI shows the numbers on every run. Module: 18 passed.


Artexis10:
**Lean suite on the final tree (`37bf42dc`, post-merge, one process, no early stop)**

`timeout 7200 .venv/bin/python -m pytest -q -p no:cacheprovider --ignore=tests/test_memorybench_export.py --ignore=tests/test_memorybench_setup.py`

`36 failed, 12939 passed, 264 skipped in 2208.33s (0:36:48)`

35 of the 36 were one cause and not this change: the lane venv still carried `exomem 0.57.2` metadata after the `origin/main` merge brought the 0.58.0 release bump, so `doctor` fails `install.lock_parity` and every test that drives the demo, doctor or setup wizard exits 1. After `uv sync --frozen` (installed 0.57.2 → 0.58.0) the seven affected files re-run as **202 passed, 1 skipped, 1 failed** — the one failure is `test_bounded_graph_join::test_a_direct_mutation_guard_also_returns_while_its_rebuild_runs`, the thread-race assertion already documented as flaky on `main` (#725). CI on a quiesced runner is authoritative for that one; the 27 checks here are the evidence that counts.

PR #759, opened 2026-08-23
fix(hosted): make the reviewer evidence window a flag, not 55 hardcoded minutes

Artexis10:
## Why

Substrate's `decouple-reviewer-assignment-expiry` (substrate#140) made the reviewer rollout
assignment take its expiry from the staged client release rather than
`LEAST(authority, stage)`. The staged release is now the **whole evidence window** — it
bounds the internal-canary credential, all four canary token paths, `storeClientArtifact`
(the `import` step), and the `cells` precondition inside `promoteExomemHostedCohort`.

This harness still hardcoded 55 minutes in both `create-stage` calls — the Claude stage in
`prepare` and the OpenAI sibling stage in `run`. That number was chosen when the authority's
thirty-minute cap won the `LEAST` anyway, so it could not matter. Now it is the entire
budget for:

- provisioning the cell (~10 minutes observed on 2026-08-22),
- a clean-client run of seven manual operations in a real client, **per platform**,
- `observe` → `sign` → `import` for each,
- `promote`.

For two platforms that does not fit. Worse, the Claude stage is created at `prepare`, so it
is already burning down before `run` is invoked.

## What changed

`--window-minutes`, default **180**, applied to both stages. `boundedExpiry` caps a staged
release at seven days, so the operator chooses within a bound the server enforces rather
than an arbitrary one.

Also `OPERATOR_CLIENT_BOUND` 32 → 96, matching substrate migration `0051`. Preflight was
reporting `<=0 of 32 operator slot(s) free` against a deployed bound of 96, which reads as a
hard blocker when 64 slots are actually free. It is a duplicate of a value the database
owns and it will go stale again — the real fix is for the control plane to report headroom
and for this to read it, tracked as task 3.2 of substrate's `reclaim-oauth-client-slots`.
The comment says so at the definition rather than leaving the next person to rediscover it.

## Verification

- `tests/test_reviewer_bootstrap_cli.py` — 28/28.
- `--help` renders the new flag; module parses.
- Behaviour is unchanged at the default for a single-platform window; only the number moves.

## Note

Pass the same `--window-minutes` to `prepare` and `run`. They create separate stages and
there is no shared state between the invocations that would carry it.

### DIFF
--- a/scripts/reviewer_bootstrap.py
+++ b/scripts/reviewer_bootstrap.py
@@ -64,9 +64,27 @@ from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
 CANDIDATE_PROFILE = "hosted-alpha-agent-v1"
-#: `exomem_oauth_client_partition_available` in migration 0048. Operator clients
-#: keep the original bound; auto-registered CIMD clients get a separate 128.
-OPERATOR_CLIENT_BOUND = 32
+#: `exomem_oauth_client_partition_available`, raised from 32 to 96 by substrate
+#: migration 0051 after two weeks of promotion attempts filled the partition and
+#: blocked the first promotion outright. Auto-registered CIMD clients keep their
+#: separate 128.
+#:
+#: This duplicates a value the database owns and will go stale again the next
+#: time the bound moves -- it already did once, reporting "<=0 of 32 free" while
+#: the deployed bound was 96. The fix is for the control plane to report headroom
+#: as server state and for this to read it; tracked as task 3.2 of the substrate
+#: change `reclaim-oauth-client-slots`. Until then, keep this in step with the
+#: latest migration to touch the function.
+OPERATOR_CLIENT_BOUND = 96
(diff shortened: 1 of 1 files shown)


Artexis10:
Superseded — this landed by another route. `main` carries the full `--stage-minutes` implementation (`DEFAULT_STAGE_MINUTES = 55`, `MIN_STAGE_MINUTES = 20`, `MAX_STAGE_MINUTES = 7 * 24 * 60`), both `create-stage` calls take it, and `run` reuses `context["stageExpiresAt"]` rather than recomputing from the CLI default. The remaining diff here is reworded comments, and the branch is now conflicted.

Worth recording why this sat open for thirteen days: its Why section correctly stated that substrate#140 moved the assignment onto the staged release's clock. `scripts/promotion_evidence.py` still said the opposite in its docstring, and that stale paragraph — not this PR — is what later sessions read. It cost a planning cycle on 2026-09-05, where a substrate change was designed to decouple an expiry that was already decoupled. The docstring is corrected in #1083.

PR #855, opened 2026-08-26
fix(governance): hide ambiguous reference candidates

Artexis10:
## Summary

- resolve canonical memory references from the read-only current corpus before public ambiguity reduction
- remove reserved, tombstoned, blocked, unresolved-principal, and L0 candidates before deciding not-found, unique, or ambiguous
- keep a visible-only ambiguity count while making hidden duplicates and stale reserved index rows identical to physical absence

## Why

Raw reference resolution counted every colliding page. Two hidden pages therefore returned AMBIGUOUS_REFERENCE instead of the same REFERENCE_NOT_FOUND envelope as absence, and one hidden duplicate could shadow an otherwise unique visible page. This closes that error/count oracle at the shared command boundary.

This is a bounded part of OpenSpec task 8.5; the broad task remains open for the other malformed/stale/parser/candidate-safety cases.

## Verification

- red first: 2 failed, 1 passed for hidden-absent, hidden-plus-visible, and visible-only collision cases
- 449 passed, 1 dependency/platform skip: governance postfilter, direct get payload, egress, and memory-reference suites
- 194 passed, 93 deselected: reference/identifier/reserved-path regression slice
- Ruff on all four changed files
- uv lock --check
- strict OpenSpec validation
- public-artifact privacy validation
- git diff --check

## Stack

Stacked on #849. Keep unmerged until the Personal Exomem stability audit clears the governance stack.

PR #856, opened 2026-08-26
fix(governance): fail closed on malformed semantic metadata

Artexis10:
## Summary

- preserve parser validity on immutable Markdown snapshots
- fail closed when malformed or unterminated frontmatter leaves semantic scope membership undecidable
- preserve conservative path/ref-only classification and make direct-read refusal identical to physical absence

Stacked after #855. This does not merge or touch any live vault.

## Verification

- red-first oracle: 2 expected failures and 1 path-only control pass before implementation
- exact regression: 4 passed
- focused governance/direct-read/parser suites: 476 passed
- find and line-ending parser suites: 49 passed
- Ruff on all six changed files
- uv lock --check
- openspec validate harden-governance-for-consolidation --strict
- git diff --check

### DIFF
--- a/src/exomem/find_corpus.py
+++ b/src/exomem/find_corpus.py
@@ -219,6 +219,7 @@ def parse_page(
     text = text.replace("\r\n", "\n").replace("\r", "\n")
 
     fm_match = FRONTMATTER_PATTERN.match(text)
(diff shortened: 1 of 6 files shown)


Artexis10:

PR #870, opened 2026-08-27
fix(governance): migrate enrolled v3 store

Artexis10:
## Summary

- derive the exact initial v4 active tuple from one validated immutable migration seed before enrollment
- migrate an enrolled exact-v3 sidecar in place under the held governance-store identity and cooperative writer guard
- require the external registry target, live standalone membership, and stable conflict-free policy workspace to match before any schema write
- replay an exact post-commit v4 result without consulting mutable workspace, while refusing copied custody, target drift, future schemas, and any extended v3/v4 schema

This is still an internal offline migration primitive. Its caller must drain old binaries and prepare verified catalog/projection evidence. It exposes no command, enables no live rollout, merges nothing, and touches no live vault.

## Verification

- red first: 7 expected failures before target/store APIs existed
- focused migration/refusal/replay suite: 9 passed
- migration/custody/policy packet on Python 3.13: 182 passed
- adjacent store/session lifecycle packet: 38 passed
- Ruff on all three changed files
- uv lock --check
- openspec validate harden-governance-for-consolidation --strict
- git diff --check

### DIFF
--- a/src/exomem/governance/schema_v4.py
+++ b/src/exomem/governance/schema_v4.py
@@ -599,6 +599,23 @@ def _seed_material(seed: MigrationSeed) -> _SeedMaterial:
     )
 
 
+def migration_target(seed: MigrationSeed) -> VerifiedActiveGovernanceState:
+    """Derive the exact initial active tuple before irreversible enrollment."""
+
+    material = _seed_material(seed)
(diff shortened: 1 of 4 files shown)


Artexis10:

PR #871, opened 2026-08-27
test(governance): freeze rich schema v3 migration fixture

Artexis10:
## Summary

- freeze the complete logical schema-v3 authority plus representative legacy handles, grants, purposes, tokens, receipt state, a pending recovery journal, and direct-source policy
- prove every ordinary v3/v4-capable store opener leaves the fixture byte-logically v3 with no v4 DDL or DML
- exercise all four pre-commit migration crash barriers and prove exact rollback, then prove enrolled migration archives legacy authority conservatively while preserving receipts/recovery evidence and publishing the first immutable policy/catalog/tuple
- complete OpenSpec task 5.6 across the preceding custody/enrollment/store-migration stack

## Verification

- red first: the new suite failed at collection because the frozen fixture module did not exist
- `89 passed`: model-reaper stability plus the rich v3 fixture, store migration, legacy store, and standalone provisioning packets
- 30/30 repeated event-driven model-reaper lifecycle checks on the updated #870 base
- Ruff on all touched and adjacent stability files
- `uv lock --check`
- public repository artifact validation
- `openspec validate harden-governance-for-consolidation --strict`
- `git diff --check`

Stacked after #870. This does not expose migration publicly, enable schema-v4 serving, merge branches, or touch any live Personal/POLLY vault.

PR #872, opened 2026-08-27
refactor(governance): share exact policy workspace mirror

Artexis10:
## Summary

- move the exact held-handle `_Governance` workspace mirror into a reusable policy primitive
- keep the live v4 authoring path on that one implementation, preserving its crash barriers and closed `complete` / `diverged` / `pending` outcomes
- require exact governance-tree owner authority before the first mutation and retain the existing identity-coordination/publication contract
- add direct replacement/addition/removal, drift-refusal, effect-callback, and unauthorized-call regressions

This is stacked after #871. It is the reusable filesystem half required by the explicit offline v4→v3 rollback coordinator; durable external rollback recovery metadata remains the next slice. It does not expose a rollback command, enable v4 serving, merge vaults, or touch any live vault.

## Verification

- red first: 3 expected failures because `policy.mirror_authoring_workspace` did not exist
- shared primitive + existing mirror crash/race slice: 8 passed, 101 deselected
- authoring snapshot / prospective compile slice: 12 passed, 89 deselected
- Ruff on all three changed Python files
- `uv lock --check`
- `git diff --check`

### DIFF
--- a/src/exomem/governance/policy.py
+++ b/src/exomem/governance/policy.py
@@ -31,14 +31,14 @@ import re
 import sqlite3
 import stat
 import time
-from collections.abc import Mapping
+from collections.abc import Callable, Mapping
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
 import yaml
 
-from .. import held_fs, memory_refs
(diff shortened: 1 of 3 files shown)


Artexis10:

PR #873, opened 2026-08-27
fix(governance): persist downmigration terminal

Artexis10:
## Summary

- persist a closed, immutable v3 recovery terminal inside the same transaction that closes v4 session authority and removes v4-only schema
- bind the terminal to the critical receipt event, reviewed plan, complete active policy/projector/catalog tuple, verified workspace/catalog parity, and exact closure counts
- refuse malformed or colliding recovery identities before writes and prove every existing crash barrier restores exact v4 state

This is stacked after #872. It supplies the atomic database terminal needed by the offline v4→v3 coordinator; the critical receipt coordinator and retry/replay wrapper remain the next slice. It does not expose a rollback command, enable v4 serving, merge vaults, or touch any live vault.

## Verification

- red first: success path failed because the recovery identity contract was absent
- downmigration-focused slice: 14 passed, 27 deselected
- complete schema-v4 store suite: 41 passed
- Ruff on both changed Python files
- uv lock --check
- openspec validate harden-governance-for-consolidation --strict
- git diff --check

PR #874, opened 2026-08-27
fix(governance): coordinate offline downmigration

Artexis10:
## Summary\n\n- write a deterministic, reviewed v4→v3 recovery plan and critical receipt intent before any mutable effect\n- require authenticated external custody plus an entirely drained/no-in-flight v4 serving membership\n- retain held store and governance-workspace identity coordination through mirror, commit, and replay verification\n- atomically close the same recovery journal with the exact v3 terminal and recover intent-only, partial-mirror, and post-commit crashes\n- refuse to backfill a missing receipt after a database effect\n\nStacked after #873. This PR does not merge or touch any live vault.\n\n## Deliberate boundaries\n\nThis is the offline coordinator, not public rollback activation. It does not boot an old v3 binary, advance the external recovery registry, enable v4 serving, merge Personal/POLLY, or operate on either live vault. Those remain later gated slices.\n\n## Verification\n\n- red-first: missing coordinator collection and missing core recovery-target contract\n- focused migration/mirror packet: 69 passed\n- Ruff on all four changed Python files\n- \n- Change 'harden-governance-for-consolidation' is valid\n- public repository artifact validation: 3,344 files / 3,412 payloads\n- source distribution and wheel build\n-

### DIFF
--- a/src/exomem/governance/schema_downmigration.py
+++ b/src/exomem/governance/schema_downmigration.py
@@ -1,0 +1,1154 @@
+"""Receipt-first, replayable offline governance schema-v4 downmigration."""
+
+from __future__ import annotations
+
+import base64
+import hashlib
+import json
+import sqlite3
+import unicodedata
+from contextlib import contextmanager
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Final
+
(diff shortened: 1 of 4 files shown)


Artexis10:

PR #875, opened 2026-08-27
fix(governance): enforce migration drain

Artexis10:
## Summary

- enroll standalone v3 migration custody as authenticated `DRAINING/schema-v3/issuance-stopped/no-in-flight` state
- refuse the first v3→v4 database effect unless every admitted replica proves that exact drain
- validate the complete active policy/catalog target after commit, then publish a signed epoch-2 `SERVING/schema-v4` successor
- keep the membership-before-control publication interval fail-closed and replay it exactly after a crash

Stacked after #874. This PR does not merge or touch any live vault.

## Deliberate boundary

This supplies authoritative migration-fence state but does not yet wire the Hosted/standalone deployment supervisor or start the actual old v3 binary. That rollout-fence/probe remains the next slice.

## Verification

- red-first: 4 expected failures for false v4-serving enrollment, absent drain refusal, absent epoch advance, and missing crash recovery
- exact migration suite: 11 passed
- adjacent custody/membership/migration/downmigration/store packet: 155 passed, 1 Windows-only DACL skip
- Ruff on all three changed Python files
- `uv lock --check`
- `openspec validate harden-governance-for-consolidation --strict`
- public repository artifact validation: 3,344 files / 3,412 payloads
- source distribution and wheel build
- `git diff --check`

PR #877, opened 2026-08-27
fix(governance): fence old schema replicas

Artexis10:
## Summary

- add a monotonic schema fence to the external writer-lease coordinator and revoke the predecessor holder in the same transaction
- make current replicas attest schema v4 on acquire/renew while treating the released pre-v4 wire shape as schema v3
- add an operator-only deployment admission probe with a credential distinct from the normal lease bearer
- exercise the actual PyPI `exomem==0.57.0` binary: it starts on v4, reads with governance DML, and authors content unless the external deployment/lease fence stops it
- prove the old binary can write again only after exact offline v4→v3 rollback and a fenced generation transition

Stacked after #875. This is the external fence and pinned-artifact proof; it does not yet wire fence transitions into the migration coordinator, mark OpenSpec 5.7/5.10 complete, or touch any live vault.

## Verification

- 323 focused lease, coordinator, CLI, v3/v4 migration, downmigration, serving-membership, pinned-old-binary, and CI-contract tests passed
- actual `exomem==0.57.0` probe passed
- configured Ruff on changed Python files passed
- repository Ruff correctness baseline (`--select F`) passed
- `uv lock --check`
- public repository artifact validation: 3,345 files / 3,413 text payloads clean
- `uv build`
- `git diff --check`

No Personal or delegated vault was opened or modified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -558,6 +558,19 @@ jobs:
           enable-cache: true
       - name: Validate public build inputs before packaging
         run: python scripts/validate-public-artifacts.py --repository
+      - name: Prove the pinned pre-v4 binary is externally fenced
+        env:
+          EXOMEM_DISABLE_EMBEDDINGS: "1"
+          EXOMEM_DISABLE_MEDIA_EXTRACTION: "1"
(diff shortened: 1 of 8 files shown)


Artexis10:

PR #888, opened 2026-08-27
docs(governance): close policy authority evidence

Artexis10:
## Summary

- close OpenSpec tasks 3.12 and 3.13 against the completed cumulative policy-authority stack
- record irreversible enrollment, forward migration, downmigration, predecessor restore, and semantic suspend/resume/undo as evidenced through PRs #869-#887
- update the progress baseline to the current main plus green cumulative PR #887

Stacked after #887. This changes only the OpenSpec evidence ledger. It does not merge, deploy, migrate, or touch any live vault.

## Verification

- cumulative PR #887: 20 successful checks, 8 expected skips, 0 failures after a failed-job-only rerun
- exact flaky reconnect test: 5/5 local CI-shaped passes plus successful CI rerun
- openspec validate harden-governance-for-consolidation --strict
- public repository artifact validation: 3,348 files / 3,416 payloads clean
- git diff --check

PR #889, opened 2026-08-27
feat(governance): transfer standalone attachment custody

Artexis10:
## Summary

- add an exclusive standalone attachment-transfer protocol with `SERVING -> DRAINING -> no-in-flight`, a target-bound detach acknowledgement, monotonic membership/control epochs, and exact crash replay
- keep attachment ownership in a fixed OS-derived host-control root signed by a separate host-local key, so copied vault custody or writer-state snapshots cannot revive the old root
- serialize drain with ordinary mutations and recheck attachment authority before session issuance commits
- transactionally invalidate imported session, grant, purpose, and token authority before a copied v4 target becomes reachable; exact transfer replay preserves new target sessions

Stacked after #888. This PR does not merge branches or touch any live vault.

## Security boundary

The standalone file protocol excludes rollback by the trusted OS owner/administrator of the fixed host-control root itself. Resisting that privileged rollback requires an external or TPM-backed monotonic authority and is not claimed here.

## Verification

- red-first regressions reproduced stale external-bundle revival, pre-drain issuance, copied-session resurrection, and complete pre-drain writer-state snapshot split-brain before correction
- `tests/test_authorization_standalone_provisioning.py`: 48 passed
- independent security re-review: GO, no P0/P1/P2 findings
- Ruff clean on the changed custody/session/schema/test files; `writer_lease.py` retains only its five pre-existing untouched findings
- changed production files compile
- `git diff --check`

### DIFF
--- a/src/exomem/governance/authorization_custody.py
+++ b/src/exomem/governance/authorization_custody.py
@@ -71,15 +71,56 @@ _CONTROL_FIELDS = frozenset(
 )
 _CONTROL_MAC_DOMAIN = b"exomem.authorization-session.control/v1"
 _ATTACHMENT_DOMAIN = b"exomem.authorization-session.attachment/v1"
+_DETACH_ACK_MAC_DOMAIN = b"exomem.authorization-session.detach-ack/v1"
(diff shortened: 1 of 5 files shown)


Artexis10:

PR #965, opened 2026-08-30
docs(openspec): propose frozen stance verification (S9)

Artexis10:
## What

Proposes **S9 of the no-nudge programme — frozen stance verification** (OpenSpec change `add-frozen-stance-verification`; spec artifacts only, no code). First occupant of the verifier tier the ratified authority-and-effects matrix admits, and the re-laning of the claim-polarity seam the constitution names non-compliant.

Claim polarity reaches the product on two paths today, both behind `EXOMEM_CLAIM_LEVEL` (default off): the synchronous write path (`corpus_aware._refine_contradictions`, which sharpens warning text and mints the `contradiction-band` advisory kind) and the asynchronous audit sweep (`audit._pair_polarity`, which already writes `meta.polarity*` enrichment) — with silent fallback to a lexical heuristic that self-describes as a stand-in, and a model chosen by an unpinned env knob. This change consolidates to ONE admitted channel:

- **New `frozen-verifiers` capability**: a verifier runs only under its opt-in gate (`EXOMEM_CLAIM_POLARITY_NLI`, default off) plus a repository pin registry (weights digest, versioned label map, green verification fixture set — no runtime configuration can add or alter a pin); anything else refuses to absence, never to a fallback wearing the verifier's name; classification-shaped input only, no vault text in instruction position; output is provenance-marked review-queue enrichment only; byte-identical degradation when absent.
- **`contradiction-queue` delta**: the shipped `meta.polarity` enrichment is re-laned — produced only by the admitted verifier (`polarity_method: "nli"`), gaining `polarity_model_digest` + `polarity_label_map_version`, bound to the `signal_version` it was computed against (stale labels drop, never serve). Heuristic-method queue labels are retired. The label never changes `signal_version`, ranking, caps, or recorded triage; it is explicitly distinct from the reader-recorded competing-alternatives pair stance.
- **`command-surface` MODIFIED delta**: the write path loses the in-path call entirely; the `contradiction-band` advisory kind is retired (its only producer was that call — it never fired on the default path), the canonical write-path-advisory requirement shrinks to two kinds, and the fate of stranded gate-on dismissals is stated (one-time resurfacing under the overlap identity).
- **Knob fate**: `EXOMEM_CLAIM_LEVEL` keeps gating the claim subsystem; `EXOMEM_CLAIM_NLI_MODEL` is retired (a set value selects nothing and is reported). Ships as a default-off `nli` extra, local-only in v1. Bench honesty: f22 stays withheld with sequence 2; precision is claimed against the verification fixtures only; byte-identity claims are scoped to the default gate-off path.

Merge after #963 (matrix ratification) — this change executes the constitution clause that PR ratifies.

## Verification

- `openspec validate --all --strict`: 167 passed, 0 failed (new change included).
- Two adversarial review rounds by a fresh zero-context reviewer over the actual diff and the shipped code. Round 1: REQUEST CHANGES (9 findings, including two HIGHs the first draft missed: the audit sweep already writes `meta.polarity` on the same path, and removing the write-path call silently retired an advisory kind named by a canonical SHALL). Round 2 verdict: **APPROVE** — quoted in a comment below.

### DIFF
--- a/openspec/changes/add-frozen-stance-verification/design.md
+++ b/openspec/changes/add-frozen-stance-verification/design.md
@@ -1,0 +1,69 @@
+# Design — frozen stance verification
+
+## D1. The admission rule is code, not posture
+
+A verifier is admitted by a registry entry pinning four things: the model name,
+the sha256 digest of its resolved weights, the label-map version it was
+verified against, and the fixture set that verified it. At load, the resolved
+weights are hashed and compared to the pin; a mismatch, a missing model, or an
+unverified (name, label-map) pair refuses the verifier. Refusal degrades to
+**absence** — the entry simply carries no stance — never to the lexical
+heuristic wearing the verifier's `method` label. The heuristic remains what it
+is today (a differently-labelled, unprivileged measurement) and is not part of
+this slice's queue enrichment.
+
+## D2. Stance is enrichment, not a new entry kind
+
+`meta.provenance` keeps its closed set (`asserted` / `proximity`). A proximity
+entry gains an optional `meta.stance` block: `label`, `method: "nli"`,
+`model_digest`, `label_map_version`, and the score as a measurement (precedent:
+the cosine already rides the entry). Three invariants keep it inside the
+matrix: attaching or changing a stance never changes `signal_version` (no
+dismissal resurfaces because labelling arrived — the same guarantee the
+provenance-labelling requirement made); ranking is untouched (asserted above
+proximity stands; stance never reorders); and canon, decisions, retrieval and
+policy never read it — it exists for the reader and the decider on the review
+surface.
+
+## D3. Invocation point: the audit's contradiction pass
+
+Enrichment runs where the queue is already assembled — the asynchronous
+audit/sweep path — bounded per pass and failure-isolated per entry (an
+exception leaves that entry unenriched and recorded as degraded, mirroring the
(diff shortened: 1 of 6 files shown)


Artexis10:
Fresh zero-context adversarial review, two rounds over the actual diff and the shipped code.

Round 1: REQUEST CHANGES — 9 findings. The two structural HIGHs: the draft's "one invocation path" premise was false (the audit sweep already writes `meta.polarity` enrichment through the same seam, so the change would have created a dual channel), and removing the write-path call silently retired the `contradiction-band` advisory kind that a canonical `command-surface` SHALL names, stranding recorded dismissals with no stated fate.

> **APPROVE** — every finding is closed with normative text plus a red-first task rather than prose, the `command-surface` MODIFIED block preserves the canonical requirement verbatim apart from the intended retirement, and the F2 consolidation onto the shipped `meta.polarity` channel eliminates the dual-channel defect at the root.

The reviewer verified the MODIFIED block programmatically against the canonical spec (all four scenarios and both retained paragraphs byte-identical; only the intended retirement altered) and re-verified each code claim against `corpus_aware.py` / `audit.py` / `claims.py`. Two residual nits recorded as non-blocking (one loose sentence in Why about the heuristic fallback; the "no-verifier build" baseline being hypothetical-but-determinate).

Merge order: #963 (matrix ratification) first — this change executes the clause that PR ratifies.

PR #966, opened 2026-08-30
docs(openspec): propose due-state carriage on the operation leaves (S6 bulk-carrier)

Artexis10:
## What

Proposes **due-state carriage on the operation leaves** (OpenSpec change `extend-due-state-to-bulk-carriers`; spec artifacts only, no code) — the S6 bulk-carrier founder decision (2026-08-30): the operation leaves carry the bounded advisory due-state block, batch-once.

One ADDED `command-surface` requirement: enumerated mutating invocations of `adopt_vault`, `adoption_studio` (`apply`/`apply-proposal` only — previews would burn the session's single change-only emission), `maintain_memory` (`fix` with `dry_run=false`, `reconcile`, `backfill-ids` with `dry_run=false`, `structured-files` with `apply=true`), `preserve_artifacts`, and `process_media` carry the block **when the invocation commits at least one governed write**, under the existing carrier contract and unchanged canonical emission governance. The no-write case (clean-vault repair, already-valid media, a retry re-enqueue) carries nothing — it has no committed terminal — and that limitation is stated, not papered over, as is the remote-surface bound (mutating maintenance other than structured-files is local-CLI-only).

One MODIFIED `command-surface` requirement: "The f23 family runs against the real runtime" — the emission assertion's recorded outcome moves from `unsupported` (measured zero carrier) to decided (one block against twelve writes). The two zero-carrier tripwire tests in `tests/test_due_state_emission_capture.py` are named for **inversion, not deletion**; design D5 adjudicates on the recorded precedent why an assertion flipping `unsupported` → decided needs no §7 bench amendment (no family, assertion, predicate, gate, or OpKind changes). f26 stays withheld with sequence 2 and gets recorded at activation.

## Verification

- `openspec validate --all --strict`: 169 passed, 0 failed.
- Two adversarial review rounds by a fresh zero-context reviewer over the actual diff and shipped code (65 tool uses, tracing the terminal path, the leaf registries, the bench journeys, and the tripwire tests). Round 1: REQUEST CHANGES — 1 CRITICAL (the change silently invalidated the f23 bench scenario and two deliberately-built tripwire tests with no MODIFIED block), 7 MAJOR (dry-run defaults, omitted maintenance modes, the no-write case, projection reality for two leaves, remote-surface bound, tautological scenario). Round 2 verdict: **APPROVE** — quoted in a comment below.

### DIFF
--- a/openspec/changes/extend-due-state-to-bulk-carriers/design.md
+++ b/openspec/changes/extend-due-state-to-bulk-carriers/design.md
@@ -1,0 +1,93 @@
+# Design — due-state carriage on the operation leaves
+
+## D1. Carriage rides the committed terminal, so the no-write case is out
+
+The block is admitted inside the terminal projection, which exists only when
+the invocation committed a governed write (`mark_active_mutation_committed`
+fires on a real replacement; the projection short-circuits otherwise). All
+five leaves reach that path when they write. When they do not write — a
+clean-vault `fix` pass, already-valid media, `process_media` `retry` (which
+re-enqueues in the machine-local job store and commits nothing) — there is no
+committed terminal to carry a block, and the leaf's raw payload passes
+through untouched. The requirement therefore scopes carriage to invocations
+that commit at least one governed write, honestly: a first-qualifying
+no-write maintenance pass stays silent even when the projection has open
+items. Closing that would be a response-contract change to non-committing
+responses — named future work, not implied.
+
+## D2. The invocation list is enumerated, not gestured at
+
+- `adoption_studio`: `apply` and `apply-proposal` only. The other six
+  registry mutations (`start`, `select`, `plan`, `cancel`, `finish`,
+  `propose`) are run bookkeeping or previews; letting a `plan` preview carry
+  would burn the session's single change-only emission before the real apply
+  — the exact failure the terminal's admission logic records having fixed
+  once.
+- `maintain_memory`: `fix` with `dry_run=false` (the mode DEFAULTS to a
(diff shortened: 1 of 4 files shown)


Artexis10:
Fresh zero-context adversarial review, two rounds over the actual diff, the shipped code, and the bench artifacts.

Round 1: REQUEST CHANGES — 1 CRITICAL, 7 MAJOR. The CRITICAL: the first draft would have silently flipped the f23 bench family's recorded `unsupported` outcome and tripped two tests deliberately built as tripwires for this exact change, with no `MODIFIED` block and no mandate — "landing the carrier without touching either isn't adding carriage, it's making three artifacts lie." The MAJORs included `maintain_memory` fix defaulting to a dry-run preview (the requirement contradicted itself), two omitted maintenance modes (including the one already terminal-wired), the unimplementable no-write case, and a headline scenario that was "the exact tautology this repo already removed from its own test suite."

> Every CRITICAL and MAJOR from the original pass is closed with text I verified directly, the MODIFIED block is a clean two-line rewrite with no drift, and the two places the change could have over-claimed (the no-write case, the remote-surface bound) it instead states the limitation up front. The residual items are cosmetic.
>
> **APPROVE**

The reviewer diffed the MODIFIED block against canonical programmatically (exactly two lines differ) and verified the new claim's mechanism (f23's block delivers via first-of-session, not a count change — noted for the driver-docstring update in tasks 2.1). The three cosmetic residuals were applied in a6f91acb except the directory name, which stays.

PR #1009, opened 2026-08-31
feat(claims)!: freeze stance verification off the write path

Artexis10:
## What changed

Stance inference no longer participates in synchronous writes. The `contradiction-band` advisory kind and its polarity clause are retired; write-time duplicate guidance remains deterministic `near-duplicate` or `overlap` output.

Optional model labels now enrich only the bounded contradiction review queue, and only after an in-repo model pin, exact resident-weights digest, versioned label map, and fixture set all agree. Loading uses the exact hashed snapshot with local-only semantics. Any missing dependency, bad digest, non-finite logits, failed fixture, stale signal, or per-entry error degrades to no label.

The repository intentionally ships an empty pin registry. A real pin, real-model CI, and the orphaned max-pairs knob decision remain explicit follow-ups; this PR does not imply a model is active.

The OpenSpec change is synchronized into the canonical specs and archived.

## Breaking change

`contradiction-band` is no longer a registered write advisory kind. Existing dismissals for that retired identity do not suppress the replacement `overlap` identity, so an affected pair may resurface once.

## Verification

- Combined affected surface after integrating current main: 387 passed, 62 expected optional-dependency or Helm skips.
- Frozen verifier and doctor post-fix: 70 passed.
- Strict OpenSpec validation: 168 passed, 0 failed.
- OpenSpec/privacy/archive gates: 63 passed.
- Locked environment parity: would make no changes.
- Public artifact scan: 3,459 repository files and 3,527 text payloads clean.
- Exact full-suite comparison against the Lane B main baseline: 98 common red nodes, 0 baseline-only, 1 branch-only, 0 kind changes; the sole branch-only node passed immediately in isolation. The subsequent main commit touched only hosted Helm files and its contract suite is included above.
- Ruff and `git diff --check`: clean.


Artexis10:
Independent Codex review: APPROVE `b1bf98b7564072d945a98a16060408dce9533aa6` with no findings.

The reviewer reproduced the targeted packaging contract (1 passed), frozen-verifier and doctor suites (70 passed), locked environment parity, strict OpenSpec validation (168 passed, 0 failed), and `git diff --check`. The follow-up dependency fix was confirmed limited to the supported lower bound, matching lock metadata, its contract test, and archived evidence.

PR #1139, opened 2026-09-08
fix(provisioner): enforce signed hosted capacity limits

Artexis10:
The provisioner could reserve six user cells even though the signed admission policy allows four. Restore the four-user and six-potential-attachment limits while preserving two recovery slots and continuation of existing reservations.

Signed-receipt regressions cover observed cells, retained reservations, concurrent final-slot requests, existing reservation continuity and agreement with both shipped policy contracts.

Validation: ordinary provisioner suite 752 passed; real PostgreSQL 17 suite 10 passed, 7 built-image-only cases skipped; hosted-platform operations 61 passed. Ruff, public-artifact validation and whitespace checks passed. The PostgreSQL row-lock concurrency test was updated and exercised against PostgreSQL 17. No deployment or capacity-policy increase is included.


Artexis10:

PR #1135, opened 2026-09-08
docs(hosted): plan private media, direct HTTPS and capacity checks

Artexis10:
Hosted Exomem needs document, image and audio preservation with useful OCR/search, bounded processing costs and verifiable privacy controls. This OpenSpec proposal defines encrypted transport and storage, tenant key custody, protected backups/exports, constrained media workers and staged CPU processing.

The direct HTTPS design uses a restricted public edge on the existing node, automatic certificates and exact endpoint agreement across OAuth, signed client artifacts and enrollment. The only existing hosted connection is an operator test connection, so switching endpoints resets test authorization and requires reconnecting; customer/session migration and legacy rollback publication are unnecessary. Vault data and custody remain protected.

The proposal records the four-user policy versus six-user reservation mismatch and separates its bounded repair from five-cell capacity evaluation. Direct latency acceptance compares equivalent authenticated workloads, separates warm/cold/streaming and real-connector paths, and retains the launch targets without promising a topology-only speedup. Paid GPU execution, expanded storage and live activation remain separately gated.

Validation: all 190 OpenSpec records pass strict validation with repository-pinned OpenSpec 1.10.0; archive discipline, public-artifact validation and whitespace checks pass. Independent security review approved the final direct-ingress, signed-artifact ownership, test-only switch and latency design. This PR changes planning artifacts only; implementation proceeds in bounded isolated tranches against refreshed launch contracts.

Reviewed implementation is available in [direct edge #1156](https://github.com/Artexis10/exomem/pull/1156), [client artifacts #1153](https://github.com/Artexis10/exomem/pull/1153), and [Substrate resource/OAuth #169](https://github.com/substrate-systems/substrate/pull/169). The bounded capacity consistency repair is [#1139](https://github.com/Artexis10/exomem/pull/1139). These PRs do not activate the direct profile or establish five-cell capacity.

The independently approved database TLS slice is [Substrate #170](https://github.com/substrate-systems/substrate/pull/170), covering both interactive PostgreSQL transaction entrypoints with real peer-identity rejection tests. The CPU document/OCR packaging slice is [Exomem #1158](https://github.com/Artexis10/exomem/pull/1158), verified with real offline Office/PDF/image extraction. Neon HTTP policy, protected worker admission and live activation remain open in the plan.

PR #1153, opened 2026-09-08
feat(hosted): add endpoint-bound direct client artifacts

Artexis10:
Adds a separate direct HTTPS candidate for Claude and OpenAI. Its repository-controlled endpoint binds the OAuth resource metadata, packages, archives and locks while the authorization issuer stays on the website. Existing candidates remain byte-identical. OpenAI validation also rejects a package whose lock names a different endpoint.

This is the artifact-producer part of OpenSpec `add-private-hosted-media` task 2.7 ([design PR #1135](https://github.com/Artexis10/exomem/pull/1135)). The candidate remains pending; this change does not activate ingress or replace the signed runtime target.

Validation: 120 artifact/command-binding tests passed, generated artifacts are current, public-artifact scan and Ruff passed, and the package built successfully. Independent security review approved the final commit after reproducing and verifying the lock-endpoint correction.


Artexis10:

PR #1194, opened 2026-09-10
feat(hosted): let the control plane's database sleep when the fleet is idle

Artexis10:
## What

Stops the hosted control plane holding its serverless PostgreSQL endpoint awake around the clock. Both routine workers gain an idle backoff, the three heartbeat schedules widen past the autosuspend window, and the budget capability learns to name this failure class.

Replaces this PR's original proposal, which was to relocate the provisioner's store into the cluster. That is rejected in the change's own text: it would put the control plane's state on a node the control plane manages, so a node loss would take the fleet and the record of the fleet together, and it does not survive a roadmap that auto-provisions nodes.

## Why

Measured on the live alpha, 2026-09-10:

| Observation | Value |
|---|---|
| Endpoint active, 1–10 September | 771,672 s of 849,600 s, 91% |
| Compute billed | 198,578 CU-seconds, 55.16 CU-hours |
| Average size while active | 0.257 CU, the 0.25 floor |
| Always-on floor | 0.25 CU x 730 h = 182.5 CU-hours/month |
| Free allowance | 100 CU-hours/month |

Five components heartbeat the database and all five landed inside the five-minute autosuspend window: both routine workers at one second, and the deletion dispatcher, capacity receipt collector and reconcile cron at one minute. The endpoint stays awake for its most frequent consumer, so this is all-or-nothing; widening any four would have changed the bill by zero.

## Notable content

- **The backoff resets on work.** A busy fleet runs at the original one-second cadence; only consecutive empty passes escalate. Worst-case pickup on an idle fleet rises to the configured ceiling, which is proportionate because the reconciler submitting most of this work moves to the same cadence and a cell provision takes minutes regardless.
- **Two cadence lookup tables and an alert contract were pinned to the one-minute literal.** They fail the Helm render rather than drift, which is how they were found. The scheduler's missed-run threshold moves with the cadence, or a healthy slower tick would alert on every run.
- **A property test, not a list.** `test_no_cronjob_schedule_sits_inside_the_database_autosuspend_window` renders the chart and rejects any CronJob at one-minute cadence, so a future job has to opt into the cost explicitly.

## Verification

Full provisioner suite: 1558 passed, 67 skipped.

Helm contract suite with real Helm 3.19.4 rendering (`HELM_BIN` set): 65 passed, including the new guard.

Durability workload contract: 3 passed. `ruff check` clean on the provisioner and the changed root tests; `ruff format --check` clean on every file this change touches.

`openspec validate idle-the-hosted-control-plane-database --strict`: valid.

Five failures in `tests/test_hosted_platform_operations.py` are the known secret-registry signer tests that fail identically on a clean main worktree; they are environmental and untouched here.

## Not in this PR

The provider-side autosuspend change from five minutes to sixty seconds, the budget preflight extension, and the post-deploy measurement remain open tasks in the change. Without the autosuspend change the widened cadences reduce duty cycle but do not yet let the endpoint suspend cleanly between ticks.


Artexis10:
Converting to draft. The relocation this proposes is the wrong remedy and I should not have reached for it first.

Two objections, both raised by the operator and both correct:

**Colocation creates a correlated failure and a bootstrap ordering problem.** The provisioner's state would live on a node the provisioner manages. Losing that node would lose the fleet and the record of the fleet at the same moment, and recovery would need the database restored before anything could be provisioned. Today that state is off-node, which is a real property being traded away for a cost saving.

**It does not survive the roadmap.** The capacity ledger and Hetzner provisioning already exist in this repo, so the fleet is meant to become more than one node. A database on node one is not where multi-node control-plane state belongs, and the migration would have to be undone at launch.

The cost driver is not placement. It is that `exomem-provisioner-worker` and `exomem-volume-worker` poll on a fixed one-second interval against a database billed by wakefulness, so the endpoint can never reach its five-minute autosuspend. The remedy that matches the cause is to stop being always-on: back the workers off when the fleet is idle, wake them in-cluster from the API when an operation arrives, and sleep until the next scheduled due time otherwise. The reconcile cron needs to clear the same window.

The `hosted-database-budget` delta in this change stays useful and mostly stands, except that its placement requirement names relocation as the remedy where it should name either relocation or removing the always-on property.

Rewriting along those lines rather than merging this.

PR #1238, opened 2026-09-13
fix(hosted): authenticate an abandoned release without a predecessor

Artexis10:
Design decision 3b of `simplify-hosted-launch-boundaries` fixed the first-consumer deadlock but deferred one thing: how a retained live provisioning attempt is resumed once the repair ships. Measuring the retained attempt answered it.

## What the measurement showed

The pre-repair worker applied the storage-only release with a wait the unconsumed claim could never satisfy. Every attempt timed out after 5m and retried, for 8h50m and 95 revisions. Two artifacts remain, and neither indicates a foreign writer:

- A provision that has never once succeeded accumulates only failed revisions, so the abandoned record is a pending *upgrade* with no successful predecessor. Abandoned-record authentication refuses that shape before reaching the recorded-target comparison that actually proves ownership.
- Resource ownership is registered only after an apply returns, so the operation went terminal owning only its namespace. The existing init-retry recovery correctly refuses that state, since it requires release, namespace, claim and volume ownership.

## What this change adds

- **Design 3c** — authenticate an abandoned record on chart identity, newest-revision position and exact recorded-target equality when no successful revision exists. Every other refusal stays: more than one pending record, a non-newest pending record, a chart mismatch, any target difference; an unreadable revision stays retryable rather than terminal. The install branch and the predecessor comparison where a predecessor exists are untouched. Then a narrow same-operation reopen for a provision owning only its namespace, with its full read-only precondition set, what it preserves, and the fixed order: deploy the repair, then reopen, resume at the non-waiting shell apply, proceed through `gpi1:binding`.
- **One ADDED requirement** in `hosted-tenant-cell`, with scenarios for the authenticate-and-clear path, the refusal set, the reopen preconditions and one-shot marker, and a failed precondition changing nothing. Written behaviourally, as no canonical spec names Helm.
- **Tasks A.2a–A.2c** — red-first tests for the measured shape, the reopen with an independent negative per precondition, then governed publish and ordered deploy before any execution.

No code changes. Nothing is executed against the retained cell by this PR.

## Verification

- `openspec validate --all --strict`: 199 passed, 0 failed
- `check_openspec_archive_discipline.py`: OK, 75 active changes, none task-complete
- `validate-public-artifacts.py --repository`: clean, 4194 files / 4306 text payloads


Artexis10:
Added A.2a's implementation on the same branch, so the contract and the guard it constrains review together. PR title changed from `docs(spec):` to `fix(hosted):` accordingly.

`_reconcile_own_pending_release`'s upgrade branch demanded exactly one deployed revision of the same chart. A first provision cannot have one — a cell whose every attempt failed retains only failed revisions — so its own interrupted record could never be cleared, and the recorded-target comparison that actually proves ownership was never reached. Where a deployed predecessor exists the comparison is unchanged; where none does, every retained revision must belong to this chart and the recorded target carries the proof. The install branch is untouched.

Tests went red first: the two relaxation-dependent cases failed precisely at the predecessor check before the change. Each refusal test now differs from the clearing test by exactly one varied condition — a second pending record, a non-newest pending record, a foreign chart in the retained history, a differing recorded target — with an unreadable revision staying retryable.

**Verification**
- `infra/provisioner/tests`: 1238 passed, 68 skipped, 2 failed, 17 collection errors. Both failures and all 17 errors reproduce identically on the unchanged tree — a scratch venv without `python-slugify`, so `src/exomem/vault.py` cannot import. Environmental, not attributable to this change.
- `test_helm_guarded_transition.py` alone: 63 passed.
- `ruff check` on both changed files: clean. Public-artifact gate: clean, 4194 files.

A.2b (the same-operation reopen) and A.2c (governed delivery, then execution against the retained cell) are not in this PR.

Artexis10:
A.2b added: the same-operation reopen.

An interruption during the storage-shell apply leaves an operation terminal owning only its namespace, because ownership is registered after the apply returns. Init-retry recovery requires all four of release, namespace, claim and volume, so it refuses that state correctly — and there was no supported way back.

**New modes** `shell-resume-preflight`, `shell-resume`, `verify-shell-resume`. The preflight proves read-only: one provision for the cell, terminal and finalized, no live claim or cell lock, matching fences and identities, the namespace ownership record and nothing else, a live namespace and provider object carrying this operation's envelope and digests, the fixed claim present and unbound with no provider volume, no binder Job/runtime/route, and a retained history under one chart whose abandoned record carries this operation's identity. Any mismatch refuses without mutation and without spending the one-shot. The reopen returns the operation to `namespace-ready` — the checkpoint immediately before the apply, naming the one resource it already owns — preserving identity, fence, request, namespace record, claim, reservation and invite.

**Two deliberate narrowings, both worth review.**

Exact target equality is *not* proven in the preflight. The computed target includes `authorizationSessionRevision`, and recomputing it would make the recovery observer perform a privileged session read it otherwise never needs. Instead the preflight proves the recorded target carries this operation's identity (`providerIdentity` plus `initOperationId`), and exact equality stays the apply's refusal under the guard. The residual cost: a target differing only outside the identity fields is discovered after the one-shot is spent.

Recovery settings gain four non-secret values — the pinned Helm binary, its version, the chart path and version. The alternative was decoding Helm's stored release form inside the observer, which would be a parallel implementation of the thing this codebase has already been bitten by three times. Runbook and pod spec updated.

**Anti-drift.** `provider_identity_values` is now one definition used by the values builder and by recovery. The abandoned-record position rules are one shared helper used by the apply path and the read-only preflight.

**Verification**
- `infra/provisioner/tests`: 1247 passed, 68 skipped. The 2 failures and 17 collection errors reproduce identically on the unchanged tree (scratch venv without `python-slugify`).
- New tests confirmed non-vacuous by mutation: removing the resumed checkpoint, the unbound-claim check, the retained-records check, or the identity comparison each makes its own test fail, and all pass again on revert.
- `ruff` clean; public-artifact gate clean; `openspec validate --all --strict` 199/0.

A.2c — governed publish, ordered deploy, then execution against the retained cell — is not in this PR.

Artexis10:
Review-fix pass pushed, addressing an independent adversarial review of the recovery diff.

**Blocking finding, fixed.** `shell-resume` could not succeed on the state it exists for. Its observer reused the init-retry recovery proof, which requires exactly one `status=deployed` Helm release record. A provision interrupted before its first successful apply has never written one, and the retained cell holds nine `failed` records and one `pending-upgrade`. The namespace, claim and provider-operation authentication is now one helper; the init-retry proof keeps its deployed-record requirement and the shell proof omits it.

**Evidence gap, fixed.** The one-shot now has a real-PostgreSQL suite matching the existing `reopen` suite: forced-trigger rollback leaves no marker, concurrent calls yield `resumed`/`already-resumed`, preserved ownership rows are byte-identical, an extra owned resource refuses before any live observation, and an unstable observation writes nothing.

**Also fixed**
- The preflight now applies the apply path's structural rules through a shared helper, so it can no longer admit a history the resumed apply would refuse after spending the one-shot.
- The reopen clears the `failure_attempts` budget. `pending_count` and the capacity-wait reason are kept as diagnostics; the runbook, design and spec delta are corrected to say so.
- The six cloned reopen constructs are consolidated behind thin wrappers.
- An untested `initOperationId` predicate now has a negative, and a marker test that could not fail is removed.

**Deliberately not adopted:** a symmetric init-retry-marker clause on `reopen`. The review's own trace shows the two states cannot overlap, and the clause would strand a shell-resumed operation that later reaches the init-retry false negative. Separately, the redundant reverse clause is dropped from `shell_resume`'s conditional update, whose resources check already proves the same thing within the transaction.

**Verification**
- Scoped provisioner suites, PostgreSQL 17 included: 177 passed, repeated three times.
- Ten mutations each fail their own test and restore byte-identically. That includes removing one guard from the shared admissibility decision, which breaks the shell suite and the init-retry PostgreSQL suite alike.
- One post-mutation run showed a single failure whose name was not captured. It did not recur in 14 further runs; it is recorded here as unexplained, not as a known flake.
- `ruff`, the public-artifact gate and `openspec validate --all --strict` (199/0) are clean.

**Before A.2c:** the operator's egress IP has changed from the firewall's single admin source, so SSH to the alpha node times out. Any live step needs a reviewed firewall change first.

Artexis10:
**Second independent review: APPROVE** — no critical or high findings. The reviewer found no observable change to init-retry behaviour from the consolidation, and could not build a counterexample for either marker-clause decision.

Its three medium findings were test and design gaps, fixed in `608e6678` without touching production code:
- The PostgreSQL suite's init-retry test double raised, and the service turns any observer exception into a refusal, so two refusal assertions could pass without exercising the guard. It now records calls, and both refusals match exact messages and assert the double was never reached.
- The statuses and refusals the consolidation turned into parameters are now pinned per wrapper.
- Design 3c no longer claims the preflight proves target equality.

From the low findings: marker-name distinctness is asserted again, request columns are pinned across the resume, and the lock deadline in the new test is five seconds (the likeliest cause of the earlier unnamed failure).

Verification: scoped suites including PostgreSQL 17 pass 180 before and after mutation. Admitting a namespace-only operation to init-retry, swapping the wrappers' statuses, and naming both markers alike each fail their test.

Deliberately left as reviewed, all low: the shell identity digest does not include the Helm history, the refusal order inside `authenticate_recovery_record` changed, and `provider_object_present` cannot be false. None affects an operator.

**Running the resume before deploying the repaired worker** would spend the one-shot on a guaranteed failure. We're relying on A.2c's deploy-then-verify order rather than adding a preflight guard. The worker now clears its own interrupted record automatically, so the manual resume should only ever be needed for the one operation that went terminal under the old guard.

**Operations:** the alpha firewall now also admits the operator's current egress address. The reviewed plan changed one resource, added one source and removed none; SSH is confirmed.

Artexis10:
Runbook fix in `be46979e`: the recovery Pod example hardcoded `EXOMEM_RECOVERY_HELM_BINARY=/usr/local/bin/helm`, a path that does not exist in the provisioner image. Helm ships at `/opt/exomem/bin/helm`, so the shell-resume observer would have failed on its first Helm call and refused with a content-free message. All four Helm settings are now read from the provisioner worker Deployment's environment, and the runbook refuses if any is empty. Checked against the live worker: `/opt/exomem/bin/helm`, `3.19.4`, `/opt/exomem/charts/cell`, `0.1.0`; the binary and chart exist in the running Pod and `/usr/local/bin/helm` does not. Docs and a test fixture only.

Artexis10:
Runbook fix in `cc3c9d05`: the shell-resume example left out `--stdin`, which the recovery helper requires before it reads the operation identity, and it also lacked the 75-second exec timeout every init-retry invocation uses. It now runs the three modes with the same wrapper, and says to read each result before running the next. It also records a point that settles the reviewer's stale-worker question: the operator Pod runs the provisioner image selected by the deployed lock, so the shell-resume modes only exist once the repaired lock has been rolled out.

Artexis10:
**Pre-execution check of the recovery Pod** — read-only, nothing mutated. This runs ahead of A.2c because two runbook bugs were already caught by reading ahead.

- **RBAC:** the `exomem-init-retry-recovery` ClusterRole grants get/list/watch on namespaces, PVCs, PVs, configmaps and pods, which covers every call the shell observer makes, including Helm's configmap-driver `history`.
- **Helm without writable homes:** the runbook Pod has a read-only root, no volumes and no Helm home variables. I reproduced those conditions in the running worker (same `/opt/exomem/bin/helm`, `HOME=/`, Helm home variables unset). Every Helm call the observer makes exited 0: `version`, `show chart /opt/exomem/charts/cell --version 0.1.0`, `history` and `get values --revision 95` on the stranded release. Nothing was written under the root filesystem, so the Pod spec needs no `/tmp` volume or Helm home variables for these read-only calls.
- **Helm settings:** the live worker injects `/opt/exomem/bin/helm`, `3.19.4`, `/opt/exomem/charts/cell` and `0.1.0`, which is what the runbook now reads.

PR #1273, opened 2026-09-14
test(hosted): drive a first hosted provision through the shipped workers on K3s

Artexis10:
## What this adds

A disposable CI drill that runs a whole first hosted provision on exact K3s. It submits the PROVISION operation a claimed invite submits and drives the shipped routine and volume-registration workers, composed by `build_live_provider_components`, until the governance migration reaches `gm1:complete`.

The existing governance drill seeds an initialized, stopped cell by hand and starts at the migration. Production has failed at seams that seeding skips: a Kubernetes-defaulted Job field, the attestation window, the capacity gate, the initializer-to-migration hand-off and the migration Job's window check. This drill produces that state through the real path, so each of those seams is crossed by shipped code against a real API server.

The test asserts the full checkpoint sequence:

`queued → namespace-ready → release-applied → gpi1:binding → gpi1:registering → gpi1:registered → gpi1:initializing → gpi1:complete → gpi1:drained → gm1:inspect → gm1:prepare → gm1:enroll → gm1:commit → gm1:complete`

It also asserts that one PVC binding carries from first consumer through the migration, that the namespace, Helm release, PVC and volume are recorded, that route closure was proven at the edge seam, and that custody ends governance-enrolled at schema 4 and DRAINING.

## How to run it

```
gh workflow run hosted-infrastructure.yml --ref <branch> -f first_provision_drill=true
```

This runs `tests/test_hosted_k3s_first_provision_drill.py` in the new `first-provision-drill` job. Locally the module skips unless `RUN_K3S_FIRST_PROVISION_DRILL_TEST=1` and `RUN_K3S_GOVERNANCE_DRILL_TEST=1` are set, since it reuses the governance drill's helpers. It also needs Docker with privileged containers and `HELM_BIN`.

## What it substitutes

Only external seams. Everything else is shipped code.

- **Database:** a disposable SQLite `OperationRepository`, with the production failure budget.
- **Hetzner volume API:** a recording stand-in for `HCloudVolumeControl`. It is the only Hetzner call on this path.
- **PV handle and location:** local-path PVs carry no CSI handle or Hetzner topology. A read-only view adds both to `read_persistent_volume` results for volume registration. Every write still reaches the real PV.
- **Capacity collector:** a receipt signed with the drill's contract key is minted from the shipped `KubernetesCapacityObserver` before every pass.
- **Public edge:** the route-closure probe reports 404, which is what an edge without routes returns. The private cell API is not reached in this phase.
- **Cluster:** the job starts its own K3s rather than the admission fixture. The kubelet registers the node as `hcloud://1`, which the capacity observer requires, and the embedded cloud controller is disabled so that ID stands. The bundled local-path provisioner backs `exomem-hcloud-encrypted-retain` with `WaitForFirstConsumer`.

## Findings it recorded

1. **The storage initializer's first attempt fails on a fresh volume.** In every run, the first init pod exits 6 with `HOSTED_OPERATOR_INTERNAL` after logging `index upsert incomplete after batch_atomic_write`. The Job's single retry reports `HOSTED_CELL_INITIALIZED`, and the provisioner accepts it.
   - **Root cause (reproduced separately):** the pod has a read-only root filesystem and no writable `/tmp`, so `tempfile.gettempdir()` fails inside the vault-creation lock.
   - **Why the retry is not a recovery:** it skips `init_vault` because the staged scaffold already exists. The published vault then lacks `_Schema/semantic-activation.yaml`.
   - **Status:** a `/tmp` emptyDir for the storage-init Job is being added in a separate change. The drill keeps such attempts as non-fatal `retriedTerminations` in its evidence and fails only when a Job reaches `Failed`.
2. **CI-built Helm fails the adapter's version pin.** `go install helm.sh/helm/v3/cmd/helm@v3.19.4` builds a CLI that reports `v3.19`. `HelmCliAdapter` requires the exact pin, so the first Helm call is refused with a provider metadata conflict. Production is unaffected, because the provisioner image installs the release archive. This job installs the same checksum-verified archive. The governance drill job still uses `go install` and will hit this at its first Helm call.

## Verification

- Dispatch run 34831843818 on `a0488dd8` (before rebasing onto #1270): `1 passed in 275.29s`, the full sequence above in 18 worker passes, one retried storage-init attempt (finding 1).
- Dispatch run 34834049829 on this head (`4c3b1357`, rebased onto #1270, which changed the migration coordinator): `1 passed in 270.58s`, the same sequence in 19 worker passes, one retried storage-init attempt (finding 1).
- Locally: pinned ruff, a cluster-free smoke of the drill's own pieces (request through the v2 model and the composed driver's target gate, submit and claim, receipt against the shipped verifier, the PV read view against `_location`), collection with and without the gates, and `scripts/validate-public-artifacts.py --repository`.


Artexis10:

PR #1333, opened 2026-09-20
feat(hosted): implement activation acknowledgement protocol and exact mutation recovery

Artexis10:
Implements OpenSpec change `repair-hosted-activation-acknowledgement`.

## Why

The first governed write on a newly provisioned hosted cell commits the catalog
and SQLite durably, then fails post-commit when `acknowledge_activation_tuple`
tries to rewrite `control.json` on a read-only mount. The store reaches epoch 2,
custody stays at epoch 1, and the request returns HTTP 409
`MUTATION_COMMITTED_ACKNOWLEDGEMENT_UNCERTAIN`. Recall, session issuance and
serving all refuse from that point, so no hosted write has ever succeeded.

## What is here

- **Exact mutation recovery** bound to canonical publication: the hosted
  mutation journal, prepared-recovery seams in the writer lease, child and
  commit components in catalog publication and the v4 schema, and
  prepare/commit/result seams across every direct publication caller.
- **Control-plane acknowledgement**: the private activation proof route, proof
  construction and verification, and the provisioner-side acknowledgement API,
  configuration and wire protocol.
- **Custody delivery**: the Unix-socket client and server, the TLS
  control-plane client, and the publisher, delivery and sidecar helpers.
- **Deployment wiring**: platform listener, ClusterIP Service and certificate
  mount; cell trust ConfigMap and egress policy; values and deployment-lock
  schema extensions; tenant and provisioner admission policy branches.
- **Certificate custody**: issuance script, startup preflight against the
  bundle the deployment lock pins, a serving re-check, and the rotation runbook.

## Crash cuts are proven, not argued

Fresh-process cuts kill a child interpreter at each barrier and reopen the vault
here. They cover the catalog publication cuts, the three Adoption Studio
completion boundaries, and the outer terminal. The cut after the authoritative
acknowledgement is the one that justifies the design: cut *before* the registry
and the next ordinary write to that page is refused
`GOVERNANCE_CATALOG_PUBLICATION_BLOCKED`; cut *after* it and the next write
simply succeeds, while a retry of the original is still refused
`STALE_SEMANTIC_WRITE` rather than published twice.

## Independent review found six defects, all fixed

Two would have stopped the worker from starting at all once the capability was
enabled — the refusal propagates out of `_run_worker`, so the pod
CrashLoopBackOffs and every cell lifecycle operation stops with it.

1. The listener's TLS Secret was mounted as a directory. kubelet publishes each
   key as a symlink into `..data/`, and the certificate reader refuses symlinks.
   Now mounted per key by `subPath`.
2. `defaultMode: 0400` on a pod with no `fsGroup` left the projection root-owned,
   denying the worker its own certificate. Now `0444`.
3. The certificate's rotation window was measured from the leaf alone, so a leaf
   could outlive the CA that signed it and nothing would say so until every
   capability-bound cell failed verification at once. Now measured against the
   chain, with the authority's expiry in the report.
4. That window was checked once per process start. A serving re-check now runs
   hourly, and never refuses.
5. The legacy-uncertain mark was suppressed on stranded cells — the ones a
   restore can revive into looking clean. Now decided from the inputs, and the
   parameter lost its default so forgetting it is a `TypeError`.
6. The floor-2 incompatibility refusal was sited on the runtime read path, where
   it cost the cell rather than the operator. Relocated to custody mount.

## Corrections to the earlier description of this work

- **There is no one-hour acknowledgement fuse.** An earlier version of this
  description claimed a pending acknowledgement blocks readiness attestation and
  strands the cell within the hour. It does not:
  `mint_hosted_replica_readiness_attestation` never compares the store's
  activation tuple to `control.json`, and renewal posts the attestation endpoint
  directly. Confirmed independently. The honest consequence of an unverifiable
  listener certificate is a fleet-wide outage, not irreversible loss. A fuse does
  exist on the restart path for a cell with an activated floor-2 vocabulary
  authority, which the alpha shape (control v1, floor 1) never reaches.
- **The vocabulary-authority collision** is recorded in `design.md` Decision 13
  and now refused at custody mount. Still latent for the alpha: cells mint at
  floor 1 and no provisioner code raises it.

## Two tasks withdrawn with evidence rather than built

- **3.4** proposed widening the readiness proof. The justification did not
  survive measurement.
- **5.4.2** would have wired the preflight classifier into the migration
  coordinator and the acknowledgement path. Each already implements its half,
  and better: the acknowledgement path validates the full predecessor-successor
  transition, and the coordinator *reissues* a short window rather than refusing.
  Wiring the classifier there would replace a repair with a refusal on precisely
  the cells `refresh_drained_source_bundle` exists to rescue. A test pins the
  conflict so it is seen before anyone tries again.

## Verification

- Provisioner corpus: 1748 passed, 69 skipped, via the invocation
  `infra/scripts/validate.sh` uses.
- Scoped: backfill/companion 143 passed 10 skipped; activation-ack and
  certificate suites 137 passed; vocabulary/admission/mount/custody 634 passed
  14 skipped; helm contract 8 passed against pinned Helm 3.19.4.
- Admission counts the tenant policy pins now verified against the rendered
  chart in both capability modes, and mutation-checked.
- `openspec validate repair-hosted-activation-acknowledgement --strict` valid.
- Public-artifact privacy gate clean on every commit.

## Why this is still a draft

Three things are outstanding and none is code.

1. **The production certificate has not been issued.** It needs the operator's
   SOPS age recipients. The matrix entries are worked out and validated but
   cannot land before the artifact exists, or the signed active registry is
   incomplete. `openspec/changes/.../tasks.md` 5.3.2 carries the exact command
   and the follow-up edits.
2. **CEL semantics and the admission acceptance/refusal run** need the desktop
   K3s. The counts are verified offline; what the expressions *mean* is not.
3. **The real mounted-image rehearsal and owner acceptance** (tasks 6.3, 6.4).


Artexis10:

